PR #12, opened 2015-02-15
Show cpu-average in /proc/stat and start cup numbering at 0

brauner:
/proc/stat mounted at /var/lib/lxcfs/ is currently missing (a) the cpu average-line from the hosts /proc/stat and (b) the numbering of the cores starts with 1 and not with the 0. (This will e.g.  cause confusion for top/htop in the container.) This patch fixes both.

Signed-off-by: Christian Brauner <christianvanbrauner@gmail.com>


stgraber:
That was merged, though apparently not through github. Closing.

PR #18, opened 2015-04-02
add cache for proc file,  for support multiple read

hustcat:
Add cache for proc file.
Save all data to cache when first read, and then return data from cache for the next read for the same opened fd.

Some tools like cat, with buffered IO, if result is too long, will cause kernel read more than once, and lxcfs will return INVAL error：

```
In container:
#strace -f cat /proc/cpuinfo
…
open("/proc/cpuinfo", O_RDONLY)         = 3
fstat(3, {st_mode=S_IFREG|0444, st_size=21786, ...}) = 0
read(3, "processor\t: 0\nvendor_id\t: Genuin"..., 32768) = 7264
write(1, "processor\t: 0\nvendor_id\t: Genuin"..., 7264processor        : 0) = 7264
read(3, 0xf81000, 32768)                = -1 EINVAL (Invalid argument)
write(2, "cat: ", 5cat: )                    = 5
write(2, "/proc/cpuinfo", 13/proc/cpuinfo)           = 13
write(2, ": Invalid argument", 18: Invalid argument)      = 18
```

In addition, cache can also support DIRECT IO.

Signed-off-by: Ye Yin <eyniy@qq.com>

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -25,6 +25,11 @@ lxcfs.1: lxcfs lxcfs.man.add
 	$(HELP2MAN) -n "Set up cgroup fs for containers" --no-discard-stderr -s 1 -I lxcfs.man.add -N ./lxcfs > lxcfs.1
 endif
 
+TEST_READ: tests/test-read.c
+	$(CC) -o tests/test-read tests/test-read.c
+
+tests: TEST_READ
+
 distclean:
 	rm -rf .deps/ \
 		INSTALL \
--- a/lxcfs.c
+++ b/lxcfs.c
@@ -43,6 +43,25 @@ struct lxcfs_state {
 };
 #define LXCFS_DATA ((struct lxcfs_state *) fuse_get_context()->private_data)
 
+#define MAX_DATA_BUF_SIZE 65536 /*64Kb should be enough*/
+
+struct file_cache {
+	size_t max_size;
+	size_t actual_size;
+	char *buf;
+};
+
+enum file_name {
+	FILE_MEMINFO   = 0,
+	FILE_CPUINFO   = 1,
+	FILE_UPTIME    = 2,
+	FILE_STAT      = 3,
+	FILE_DISKSTATS = 4,
+	NO_USED_FILE= 5, 
+};
+
+struct file_cache caches[NO_USED_FILE];
+
 /*
  * TODO - return value should denote whether child exited with failure
  * so callers can return errors.  Esp read/write of tasks and cgroup.procs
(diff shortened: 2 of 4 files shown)


hallyn:
Thanks - the commit looks very nice, except for what seems like one underlying
assumption, which is that it is ok to cache these globally.  But if two tasks
are doing interleaved partial reads,s

         P1                          P2
 open(/proc/meminfo)
                             open(/proc/meminfo)
 read two bytes
                             read two bytes
 read two bytes
                             read two bytes
...

then if P1 and p2 have different cgroup-enforced memory limits, both P1 and
P2 will see P2's contents (after the first two bytes) right?

This would be correct if the global cache was removed, and we instead kept
a structure in memory, created at FUSE fd open, and freed at FUSE fd close,
which contained this buffer.  That very much needs to be done anyway to
reduce our current overhead a bit.

Would you have time to add that to this patch?  (If not I'll try to get to it
asap and then you can build this patch on top of it.)

hustcat:
@hallyn I'd like to wait for you design some data structure for each FUSE fd, because I do not have much time.
In practice, I run one lxcfs for each container. I run lxcfs with buffer IO before([see here](https://bitbucket.org/hustcat/kernel-2.6.32/src/0e5d90ed6f3ef8a3b5fe62a04cc6766a721c70f8/fs/fuse/file.c?at=master#cl-2410)).
But I found some problems:
With buffered IO, you can't get different result for the same inode because kernel's page cache.So I run one lxcfs for each container. Howerver,after that, I found another problem. I can't get result real time because kernel's page cache. Only with `auto_inval_data` flage, kernel will invalidate inode's page cache([here](https://bitbucket.org/hustcat/kernel-2.6.32/src/0e5d90ed6f3ef8a3b5fe62a04cc6766a721c70f8/fs/fuse/inode.c?at=master#cl-222)). However, fuse userspace library still not support this flag now.

To solve these problems, I run lxcfs with direct IO([here](https://bitbucket.org/hustcat/kernel-2.6.32/src/0e5d90ed6f3ef8a3b5fe62a04cc6766a721c70f8/fs/fuse/file.c?at=master#cl-2429)):
```
./lxcfs -o direct_io ...
```
We must use `-o direct_io` also, if we run only one lxcfs for multiple containers.

hallyn:
Hi - the branch called 'cache.1' at github.com/hallyn/lxcfs (commits at https://github.com/hallyn/lxcfs/commits/cache.1 ) does a basic implementation of caching some file info at opendir and open.

At this point you should be able to use the file_info->buf and file_info->bufsize (both initialized to 0 and currently unused) from the fi->fh to store the information you wanted to cache.

hustcat:
@hallyn I create another PR [#22](https://github.com/lxc/lxcfs/pull/22)

PR #21, opened 2015-04-07
direct_io patch

q53:


hallyn:
Hi,

would you prefer for me to rewrite this patch, or would you rather do it yourself?

If you'd like to do it, then please do not fail at startup when old options are given.  So simply check for and discard '-s', '-f', and '-o allow_other'.

The 's/EINVAL/0' in the proc_*_read when offset != 0 can simply be dropped.  The one in cg_read I'm still not sure about, but it seems to be working ok.

hallyn:
Hmm.  Actually with your original tree, starting a vivid container causes lxcfs to crash

q53:
M-m-m., the patch is not intended to include as is, it is just a demo shows the conception  works.)
Regarding to 's/EINVAL/0'.  When  a read call run over a unbuffered  file the system(kernel fuse driver?) looks like not aware a file size and do a whole page request. When a read returned only the generated data size it try to resume with  offset = generated data (a file size), and as in your code a request with offset >0 always return EINVAL, any read always ends with EINVAL.I fixed this a fast straightforward and incorrect  way (returning 0) :).  My guess is  in case of a request  with a offset >= 0 (what is absolutely legitimate) a read must run a normal way, count a length of a generated data, in case of offset < length (file size) return data[offset]...data[ (length - offset)||size ] and in case offset >= length return 0, not EINVAL. Actually it must be done for work in any IO mode and i have no a clue why it is not. Currently, in proc_cpuinfo_read(), for example, there are no checks even for  (total_len > size) and it is easy to hit a buffer overflow in case of small read requests.

lxcfs.c
```diff
@@ -514,7 +514,7 @@ static int cg_getattr(const char *path, struct stat *sb)
 		sb->st_nlink = 1;
 		sb->st_uid = k->uid;
 		sb->st_gid = k->gid;
-		sb->st_size = get_file_size(controller, path1, path2);
+		sb->st_size = 0;
```
hustcat: We can't return 0. FUSE filesystem is diferent from proc/sys filesystem which are based on [seq_file](https://www.kernel.org/doc/Documentation/filesystems/seq_file.txt).If return 0, kernel will return nothing.

hustcat:
I think [#22](https://github.com/lxc/lxcfs/pull/22) can solve @q53 's problem.

q53:
hallyn
Regarding to the segfault. I made a mistake in agv_new handling, the first arg was eaten as a program name, so lxcfs run multithreded and segfaulted in libnih. I had checked lxcfs by launch a single bash, not an entire container, and it was fine. The fix in my branch. That not discard my notes about wrong offset handling. 

hustcat
You are wrong. "kernel" returns nothing not because of proc/sys file type, but because of buffered FS access.
In first terminal:
```
[root@sl7test _lxcfs]# cat /usr/sbin/lxc.mount.hook 
#!/bin/sh
if [ -d /var/lib/cgmanagerfs/proc/ ]; then
   for entry in /var/lib/cgmanagerfs/proc/*; do
       mount -n  --bind $entry ${LXC_ROOTFS_MOUNT}/proc/$(basename $entry)
   done
fi

[root@sl7test _lxcfs]# cat /var/lib/lxc/test-3/config 
# Template used to create this container: /usr/share/lxc/templates/lxc-debian
# Parameters passed to the template: -a x86_64
# For additional config options, please look at lxc.container.conf(5)
lxc.network.type = veth
lxc.network.hwaddr = 00:16:3e:73:fb:47
lxc.network.flags = up
lxc.network.script.up = /etc/lxc/ovsup
lxc.network.script.down = /etc/lxc/ovsup
#lxc.hook.mount =/root/lxcfs/cgmanagerfs.hook
lxc.hook.mount = /usr/sbin/lxc.mount.hook
lxc.rootfs = /var/lib/lxc/test-3/rootfs

# Common configuration
lxc.include = /usr/share/lxc/config/debian.common.conf

# Container specific configuration
lxc.mount = /var/lib/lxc/test-3/fstab
lxc.utsname = test-3
lxc.arch = amd64
lxc.cgroup.memory.limit_in_bytes=8096M
lxc.cgroup.memory.memsw.limit_in_bytes=11020M

[root@sl7test _lxcfs]# ./lxcfs /var/lib/cgmanagerfs
```
switch terminal:
```
[root@sl7test ~]# free
              total        used        free      shared  buff/cache   available
Mem:       16269540      484924    13489248       33532     2295368    15404324
Swap:       8265724           0     8265724
[root@sl7test ~]# lxc-start  -n test-3 -d 
[root@sl7test ~]# lxc-attach -n test-3 /bin/su -
root@test-3:~# free
             total       used       free     shared    buffers     cached
Mem:       8290304      48256    8242048          0          0        724
-/+ buffers/cache:      47532    8242772
Swap:      8265724          0    8265724
root@test-3:~# ls -ld /proc/meminfo 
-r--r--r--. 1 root root 0 Apr 11 22:08 /proc/meminfo
```

hallyn:
@hustcat, @q53,

I've pushed patches to master which should implement this.  Coudl you both please test from git and make sure there are no regressions for you?

q53:
It  works for me.
But, again, about EINVAL:

q53:
Another strange artifact)
```
[root@chef-1 ~]# wc -c  /proc/vmstat 
2355 /proc/vmstat
[root@chef-1 ~]# wc -c  /proc/meminfo 
0 /proc/meminfo
```

hallyn:
Hi - would you mind submitting a patch for the -EINVAL on > size error?

hustcat:
proc filesystem is different from disk filesystem.Generally speaking, in practice ,we read all data sequentially, and there is nothing meanings for calling lseek in proc filesystem.

PR #48, opened 2015-10-28
Change uptime format

bmiklautz:
/proc/uptime has the format %lu.%02lu %lu.%02lu. The format used by
lxcfs doesn't consider the nano second portion of the uptime which might
cause programs that rely on that format to fail.

This commit adapts the uptime format to match the kernels by adding
trailing dummy values (.0) to the uptime and idle time values.
The parsing of /proc/uptime was updated.

Signed-off-by: Bernhard Miklautz <bernhard.miklautz@shacknet.at>

### DIFF
--- a/lxcfs.c
+++ b/lxcfs.c
@@ -2387,16 +2387,18 @@ out:
 	return answer;
 }
 
-static long int getprocidle(void)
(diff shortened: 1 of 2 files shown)


hallyn:
Looks good - thank you!

PR #49, opened 2015-10-28
fix corner cases in uptime and diskstat read

### LINKED ISSUE #33: Dash read built-in behavior change with 0.9 vs 0.7
In 0.7 when running the command: read b c < /proc/uptime
variables $b and $c are populated with appropriate data.  In 0.9 they do not get data.  When running strace on lxcfs outside the container it shows that while in 0.7 they get data sent to them in 0.9 they are getting /0 sent instead.

When I test using bash (same command above) or using cat /proc/uptime as a placebo check they both act properly in 0.7 and 0.9 both giving data for uptime.

I'm putting this ticket in at Serge's request after posting on the mailing list.  Per his note it appears to be related to the implementation of direct_io which was done before the 0.8 release.

https://github.com/lxc/lxcfs/commit/7253e0a4b84fce836a2585ee5a982311163b6757

A strace on cat is trivial but one on dash is not (read is a builtin and strace won't do builtins) so I did that one for comparison sake.

Method used
stty -echo
cat | strace dash > /dev/null
Once that is done I had to type the command I wanted to strace blind when it paused waiting for input.  I enabled echo after it was all done.

https://pastebin.com/5T1xbHDH

hallyn:
Closes #33 

The code for these (a shortcut version of the other proc_*_read ones) was
doing snprintf(buf, size, ...).  If the user only requests one byte, we
just served them a trailing 0.  Fix that.

We weren't handling reads with offset in these cases either.  Fix that.

### DIFF
--- a/lxcfs.c
+++ b/lxcfs.c
@@ -55,6 +55,7 @@ struct file_info {
 	char *buf;  // unused as of yet
 	int buflen;
 	int size; //actual data size
+	int cached;
 };
 
 /* reserve buffer size, for cpuall in /proc/stat */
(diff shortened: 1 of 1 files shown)


stgraber:
Looks good to me.

PR #51, opened 2015-10-31
Implement SwapTotal and SwapFree support for /proc/meminfo

tssge:
This patch implements SwapTotal and SwapFree support for /proc/meminfo so that programs like htop and top can see the correct swap values inside an LXC container.

In order to show the correct values kernel needs to have swapaccount=1. If however this is not the case, we will not modify the values in any way showing the host's swap values for the container.

I have tested this patch on Ubuntu 15.10 and it is working as expected.

### DIFF
--- a/lxcfs.c
+++ b/lxcfs.c
@@ -1883,8 +1883,10 @@ static int proc_meminfo_read(char *buf, size_t size, off_t offset,
 	struct fuse_context *fc = fuse_get_context();
 	struct file_info *d = (struct file_info *)fi->fh;
 	char *cg;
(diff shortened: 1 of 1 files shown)


stgraber:
Hi,

Apparently your branch is conflicting with current lxcfs master, can you rebase it?

tssge:
You're right, I was quite a few commits behind. Rebased it and it seems all checks have passed now.

hallyn:
Sorry, i will take a look at this in the next few days.

hallyn:
Thanks again, applying by hand.

PR #53, opened 2015-11-12
Add a hook to make reboot sleep 0.5s

hallyn:
To ensure that the new container does not get cached directory entry
information from before we removed the container's cgroup using cgroupfs.


stgraber:
Shouldn't there be a matching change to configure.ac and/or Makefile.am?

PR #64, opened 2016-01-01
update README about the hooks configuration needed

evgeni:
- fix the paths for the files and config
- include lxc.mount.auto = cgroup:mixed


hallyn:
Thanks!

PR #70, opened 2016-01-31
cgroup idx not exceed 30 safety check interferes with normal operation

smemsh:
in store_hierarchy(), /proc/self/cgroup is parsed up but if the idx number exceeds 30 we bomb out, thinking it's corrupt.  However on my system these are over 700:
```
$ cat /proc/self/cgroup
728:perf_event:/
727:net_prio:/
726:net_cls:/
725:freezer:/
724:devices:/
723:cpuset:/
722:cpuacct:/
721:cpu:/
720:blkio:/
```

this is probably an uncommon situation resulting from the experimentation i'm doing with cgroups -- it seems every time the cgroup is created the index number increases -- but it really should work nonetheless (and does run fine if I remove the check), there's nothing wrong/corrupt about my system in this state.  Indeed, the value could be in the thousands probably or more and still be valid.

### DIFF
--- a/cgfs.c
+++ b/cgfs.c
@@ -141,7 +141,7 @@ static bool store_hierarchy(char *stridx, char *h)
 	int idx = atoi(stridx);
 	size_t needed_len = (idx + 1) * sizeof(char *);
 
-	if (idx < 0 || idx > 30) {
+	if (idx < 0) {
 		fprintf(stderr, "Error: corrupt /proc/self/cgroup\n");
 		return false;
 	}


hallyn:
Hm, that's yucky - that'll probably mean the way those are stored
will need to change a bit.  Thanks for the report.

hallyn:
Ok so simply allocating the thousands of entries when you only have 10-12 real entries isn't optimal.  I don't think there is any reason to tie the array index to the subsystem index, so I think I'll fix it for you that way.  Now if someone does 'for i in `seq 1 2000` do mkdir /mnt/cg$i; mount -t cgroup -o none,name=cg$i cg$i /mnt/cg$i' they'll have a problem, but it takes privilege on the host to do that even with cgroup namespaces, so i think we can worry about that later.

smemsh:
Just keeping the index separate as you suggest should work fine... it might be limitation in future if more kinds of subsystems/controllers are written, but right now it probably would be fine to keep subsystem array size 32 or something and store the kernel index.

I seem to remember coming across a limitation of 32 some time ago, but cannot find it in the code at the moment, maybe it used to exist but no more.

smemsh:
sorry, I could not figure out how to withdraw my workaround while leaving the issue open.

PR #78, opened 2016-02-05
cleanups, deduplication and liblxcfstest installation

Blub:
The first 3 commits are pretty straight forward. As for the last one I don't know whether this kind of reorganization is desired by you at all. Just let me know.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -27,7 +27,8 @@ liblxcfstest_la_LDFLAGS = $(AM_CFLAGS) -shared \
 noinst_HEADERS = bindings.h
 
 sodir=$(libdir)
-lib_LTLIBRARIES = liblxcfs.la liblxcfstest.la
+lib_LTLIBRARIES = liblxcfs.la
+EXTRA_LTLIBRARIES = liblxcfstest.la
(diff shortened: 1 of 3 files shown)


hallyn:
Thanks very much, these look great on the whole.  The 'return NULL' fixes are important enough to release 0.19 as soon as we get the reload test failure figured out.

hallyn:
Note that doing

gcc -shared  -fPIC -DPIC  .libs/liblxcfs_la-bindings.o .libs/liblxcfs_la-cpuset.o   -lpthread  -pthread -O2 -pthread -Wl,-soname -Wl,liblxcfs.so.0   -pthread -Wl,-soname -Wl,liblxcfs.so.0 -o .libs/liblxcfs.so.0.0.0

by hand creates what we need for the reload test to succeed, but I can't find a simple 'make' target to do that for us, which would be much nicer.

hallyn:
Actually that bit is going to change anyway.   Can you drop just that patch for now?  Or I can merge and then revert the patch.

hallyn:
Merged by hand - thanks!

PR #85, opened 2016-02-18
Use clone() instead of fork() when creating children in a different PID namespace

Fabian-Gruenbichler:
As discussed in "Potential deadlock with lxcfs and lxc-freeze" on [lxc-dev], this replaces occurrences of the setns() - fork() pattern with setns() - clone() to avoid an [issue in glibc](https://sourceware.org/bugzilla/show_bug.cgi?id=15392), whose fork code base does not deal with PID namespaces correctly (or at all).

I am not quite sure about the usage of alloca to allocate the stack for the child processes - I copied it from lxc's [lxc_clone](https://github.com/lxc/lxc/blob/master/src/lxc/namespace.c#L49) function, but it seems potentially dangerous to allocate a stack on the stack.. There is also an ifdef for ia64 selecting a different clone wrapper in lxc's lxc_clone, which might be applicable for lxcfs as well.

### DIFF
--- a/bindings.c
+++ b/bindings.c
@@ -855,19 +855,25 @@ bool is_child_cgroup(const char *controller, const char *cgroup, const char *f)
 static bool recv_creds(int sock, struct ucred *cred, char *v);
 static int wait_for_pid(pid_t pid);
 static int send_creds(int sock, struct ucred *cred, char v, bool pingfirst);
+static int send_creds_clone_wrapper(void *arg);
 
 /*
- * fork a task which switches to @task's namespace and writes '1'.
+ * clone a task which switches to @task's namespace and writes '1'.
  * over a unix sock so we can read the task's reaper's pid in our
  * namespace
+ *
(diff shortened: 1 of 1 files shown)


hallyn:
Thanks very much.

On ia64, I don't know that we care about supporting it.  If we do, then whether we need that ifdef depends on whether libc has matured since 2008.

On alloca, it's fine for a single use, but I worry about its use in the loop in pid_ns_from_wrapper().  Perhaps in any loop run other than the first we should alloc+free?  Or maybe looping there is not really worthwhile and we should just return an error if there is an error.

hallyn:
I think I will take your patch as is and add one more patch which removes the loop.  If you disagree, please let me know (and we can proceed with new pull requests).  Thanks!

Fabian-Gruenbichler:
Removing the loop seems okay to me (also makes both directions more consistent). Thanks!

PR #98, opened 2016-03-19
Fix lxcfs mount handling with newer lxc

stgraber:
Signed-off-by: Stéphane Graber <stgraber@ubuntu.com>


stgraber:
@hallyn I have uploaded a new lxcfs to Ubuntu with this fix after testing it briefly on an affected wily system.

hallyn:
You are explicitly avoiding cgmanager, should cgmanager.lower also get a special case?

stgraber:
So I thought about it then came to the conclusion that cgmanager.lower should never exist at that point in time.

It's the container post-startup which moves cgmanager to cgmanager.lower before starting cgproxy.

hallyn:
But it is the cgproxy, not lxc, which creates .lower.  It's sort of orthogonal to how lxc feels like running right now.

hallyn:
Thanks.

PR #100, opened 2016-03-21
upstart: change how we decide to not run in a container

hallyn:
Otherwise lxcfs keeps lxc from hitting state 'started', which
in turn blocks reboot/shutdown.


hallyn:
@stgraber - does this look ok?

stgraber:
Looks good to me.

PR #102, opened 2016-03-22
implement access(2)

hallyn:

### DIFF
--- a/bindings.c
+++ b/bindings.c
@@ -1744,7 +1744,6 @@ int cg_open(const char *path, struct fuse_file_info *fi)
 		goto out;
 	}
 	if (!fc_may_access(fc, controller, path1, path2, fi->flags)) {
-		// should never get here
 		ret = -EACCES;
 		goto out;
 	}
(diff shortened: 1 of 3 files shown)


bindings.c
```diff
@@ -1770,6 +1770,60 @@ int cg_open(const char *path, struct fuse_file_info *fi)
...
+
+	pid_t initpid = lookup_initpid_in_store(fc->pid);
+	if (initpid <= 0)
+		initpid = fc->pid;
+	if (!caller_may_see_dir(initpid, controller, path1)) {
+		ret = -ENOENT;
+		goto out;
+	}
+	if (!fc_may_access(fc, controller, path1, path2, mode)) {
+		// should never get here
```
stgraber: Why not?

Can't there be any case where a user doesn't have the right permissions?

stgraber: Also, shouldn't that be EPERM?

stgraber:
This is going to need a lxcfs restart to take effect, we should have the packaging do a one-time reboot-required write for this.

hallyn:
On Mon, Mar 21, 2016 at 06:57:42PM -0700, Stéphane Graber wrote:
> This is going to need a lxcfs restart to take effect, we should have the packaging do a one-time reboot-required write for this.

Yeah  - I'm not happy about it, and technically it's not required unless
the user needs the fix for nested lxc.  Old lxcfs works fine with the
new bindings, just doesn't call the new *_access helpers.

stgraber:
Ok, so probably fine not to set the flag.

bindings.c
hallyn: No, should be EACCESS per the access(2) manpage.

hallyn: I'm actually not sure why we should never get there :)  that was cut-pasted from cg_open, and there I think that may be a legacy comment from cgmanager days.  Sitill trying to remember

hallyn:
(removed that comment)

PR #108, opened 2016-04-25
set mode 755 for $RUNTIME_PATH/lxcfs/controllers

brauner:
Let's set it to the same mode as /sys/fs/cgroup. Also, tools like Docker will
complain about not being able to stat /run/lxcfs/controllers when mode is set to
700.

(This should take care of docker/docker#20548.)

### DIFF
--- a/lxcfs.c
+++ b/lxcfs.c
@@ -799,7 +799,7 @@ static bool umount_if_mounted(void)
 
 static bool setup_cgfs_dir(void)
 {
-	if (!mkdir_p(basedir, 0700)) {
+	if (!mkdir_p(basedir, 0755)) {
 		fprintf(stderr, "Failed to create lxcfs cgdir\n");
 		return false;
 	}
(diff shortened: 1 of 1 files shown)


brauner:
To be a little more precise: Not setting /run/controllers to 755 will prevent docker from running containers. (That is lxc/lxd due to lxcfs and docker cannot play together on the same host.)

hallyn:
Hm.  Not a fan of this.  This seems like a docker bug.  If we change these to 755 then next thing docker will want to use them for cgroup access, right?

In cgmanager we mounted these in a private namespace so as to keep the rest of the system from messing with them.  That caused some nuisances which were finally all ironed out, but which were not trivial.  (In the end cgmanager had to pivot into a hand-constructed minimal root, which is fragile to say the least)  Maybe we need to go that route again...  But really i see no reason why Docker should be trying to use things marked as private.

brauner:
Yeah, maybe. Switching docker and lxc to unprivileged on the same host is not possible without some changes. Maybe I can look into docker for this if I find the time.

PR #109, opened 2016-04-25
bugfixes

brauner:
- Fix do_mount_cgroups(): It previously returned ret uninitialized on failure.
- Quite a few snprintf() call used size_t variables but then checked whether
  size_t < 0. Since size_t is unsigned these checks were always true. Let's use
  ssize_t instead which is signed.
- Use additional ssize_t variable to catch snprintf() error for swap
  calculation and add the value to the final result afterwards instead of
  directly.

### DIFF
--- a/bindings.c
+++ b/bindings.c
@@ -2865,7 +2865,7 @@ static int read_file(const char *path, char *buf, size_t size,
 		return 0;
 
 	while (getline(&line, &linelen, f) != -1) {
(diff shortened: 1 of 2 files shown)


hallyn:
Thanks!

brauner:
Something I wanted to point out for future reference: In the commit message I state that "Since size_t is unsigned these checks were always true." this is obviously nonsense and should read "Since size_t is unsigned these checks were always false."

PR #121, opened 2016-07-09
skip empty entries under /proc/self/cgroup

brauner:
If cgroupv2 is enabled either alone or together with legacy hierarchies
/proc/self/cgroup can contain entries of the form:

        0::/

This will cause lxcfs to fail the cgroup mounts because it parses out the empty
string "" and later on passes it to mount(). Let's skip such entries.

### DIFF
--- a/bindings.c
+++ b/bindings.c
@@ -3951,6 +3951,14 @@ static void __attribute__((constructor)) collect_subsystems(void)
 			goto out;
 		*p2 = '\0';
 
+		/* With cgroupv2 /proc/self/cgroup can contain entries of the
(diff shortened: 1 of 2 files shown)


bindings.c
```diff
@@ -3951,6 +3951,9 @@ static void __attribute__((constructor)) collect_subsystems(void)
 			goto out;
 		*p2 = '\0';
 
+		if (!strcmp(p, ""))
```
hallyn: Can you add a brief comment here (and identical at the next) explaning why, for future debuggers?

brauner: Sure.

Issue #123, opened 2016-07-17
Use already mounted cgroups


alehaa:
As far as I known from inspecting the code, lxcfs mounts all cgroups at ``/run/lxcfs/controllers``, even if the cgroups are already mounted at ``/sys/fs/cgroup``. Libvirt doesn't like this at my server (Debian jessie), and I don't understand why this is necessary.

So I checked to use the cgroups under ``/sys/fs/cgroup``:
```diff
diff --git a/bindings.c b/bindings.c
index 8186ca9..a351b9a 100644
--- a/bindings.c
+++ b/bindings.c
@@ -762,9 +762,9 @@ bool cgfs_get_value(const char *controller, const char *cgroup, const char *file
        if (!tmpc)
                return false;
        /* basedir / tmpc / cgroup / file \0 */
-       len = strlen(basedir) + strlen(tmpc) + strlen(cgroup) + strlen(file) + 4;
+       len = strlen("/sys/fs/cgroup") + strlen(tmpc) + strlen(cgroup) + strlen(file) + 4;
        fnam = alloca(len);
-       snprintf(fnam, len, "%s/%s/%s/%s", basedir, tmpc, cgroup, file);
+       snprintf(fnam, len, "%s/%s/%s/%s", "/sys/fs/cgroup", tmpc, cgroup, file);
 
        *value = slurp_file(fnam);
        return *value != NULL;
```
This gave me no errors inside the container or at the console by lxcfs. Even better: if I unmount the ones under /run/lxcfs, libvirt will be happy, too.

I'd like to know, if a patch for detecting already mounted cgroups is welcome, of if this is a scenario that should not be used by lxcfs.

brauner:
Unfortunately, we cannot unconditionally disable the cgroup mounts that `lxcfs` does under `/run/lxcfs/controllers`. `lxcfs` uses the mounts under `/run/lxcfs/controllers` to e.g. allow for things like only providing a partial view of the cgroup hierarchy. This is at least the case on systems with a kernel that does not support cgroup namespaces. Furthermore, on non cgroup namespace supporting kernels the cgroup part of `lxcfs` is needed to run systemd based containers. There was however, discussion of adding an option to `configure` to disable the cgroup part so that we can take it out on cgroup namespace enabled kernels.

alehaa:
Okay I understand. Can this be checked in any way? Then I would spend some work on this ...

[...]

PR #124, opened 2016-07-27
move lxcfs to minimal chroot

brauner:
lxcfs now creates a private mount namespace (actually rather a minimal chroot)
to hide its cgroup mounts under BASEDIR/ from other processes that would get
confused by it. However, the fuse mount usually placed under /var/lib/lxcfs (or
whatever the user gives us via argv[1]) needs to be visible in the initial
namespace. Hence, we place these mounts in different namespaces. ~~This requires
some coordination. fuse_mount() needs to be called in the initial namespace.
Afterwards we can unshare(CLONE_NEWNS), setup the cgroup mounts and create our
minimal chroot. On failure we need to switch back to the initial mount namespace
for fuse_umount() to succeed.
In addition, when we are asked to reload our dynamic library, we also need to
switch to the initial mount namespace.~~

### DIFF
--- a/lxcfs.c
+++ b/lxcfs.c
@@ -789,8 +789,8 @@ static bool mkdir_p(const char *dir, mode_t mode)
 
 static bool umount_if_mounted(void)
 {
-	if (umount2(basedir, MNT_DETACH) < 0 && errno != EINVAL) {
-		fprintf(stderr, "failed to umount %s: %s\n", basedir,
+	if (umount2(BASEDIR, MNT_DETACH) < 0 && errno != EINVAL) {
+		fprintf(stderr, "failed to umount %s: %s\n", BASEDIR,
 			strerror(errno));
 		return false;
 	}
@@ -799,7 +799,7 @@ static bool umount_if_mounted(void)
 
 static bool setup_cgfs_dir(void)
 {
-	if (!mkdir_p(basedir, 0700)) {
+	if (!mkdir_p(BASEDIR, 0700)) {
 		fprintf(stderr, "Failed to create lxcfs cgdir\n");
 		return false;
 	}
@@ -807,7 +807,7 @@ static bool setup_cgfs_dir(void)
 		fprintf(stderr, "Failed to clean up old lxcfs cgdir\n");
 		return false;
 	}
-	if (mount("tmpfs", basedir, "tmpfs", 0, "size=100000,mode=700") < 0) {
+	if (mount("tmpfs", BASEDIR, "tmpfs", 0, "size=100000,mode=700") < 0) {
 		fprintf(stderr, "Failed to mount tmpfs for private controllers\n");
 		return false;
 	}
@@ -820,9 +820,9 @@ static bool do_mount_cgroup(char *controller)
 	size_t len;
 	int ret;
 
-	len = strlen(basedir) + strlen(controller) + 2;
+	len = strlen(BASEDIR) + strlen(controller) + 2;
 	target = alloca(len);
-	ret = snprintf(target, len, "%s/%s", basedir, controller);
+	ret = snprintf(target, len, "%s/%s", BASEDIR, controller);
 	if (ret < 0 || ret >= len)
 		return false;
 	if (mkdir(target, 0755) < 0 && errno != EEXIST)
--- a/lxcfs.c
+++ b/lxcfs.c
@@ -8,28 +8,28 @@
 
 #define FUSE_USE_VERSION 26
 
-#include <stdio.h>
 #include <dirent.h>
+#include <dlfcn.h>
+#include <errno.h>
 #include <fcntl.h>
 #include <fuse.h>
-#include <unistd.h>
-#include <errno.h>
-#include <stdbool.h>
-#include <time.h>
-#include <string.h>
-#include <stdlib.h>
 #include <libgen.h>
-#include <sched.h>
 #include <pthread.h>
-#include <dlfcn.h>
-#include <linux/sched.h>
-#include <sys/socket.h>
-#include <sys/mount.h>
-#include <sys/epoll.h>
+#include <sched.h>
+#include <stdbool.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+#include <time.h>
+#include <unistd.h>
 #include <wait.h>
+#include <linux/sched.h>
+#include <sys/epoll.h>
+#include <sys/mount.h>
+#include <sys/socket.h>
 
-#include "config.h" // for VERSION
 #include "bindings.h"
+#include "config.h" // for VERSION
 
 void *dlopen_handle;
 
--- a/lxcfs.c
+++ b/lxcfs.c
@@ -953,7 +953,7 @@ static bool do_mount_cgroups(void)
 {
 	char *target;
 	size_t clen, len;
-	int i, fd, ret;
+	int i, ret;
 
 	for (i = 0; i < num_hierarchies; i++) {
 		char *controller = hierarchies[i];
@@ -977,13 +977,11 @@ static bool do_mount_cgroups(void)
 			return false;
 		}
 
-		fd = open(target, O_DIRECTORY);
-		if (fd < 0) {
+		fd_hierarchies[i] = open(target, O_DIRECTORY);
+		if (fd_hierarchies[i] < 0) {
 			free(target);
 			return false;
 		}
-
-		fd_hierarchies[i] = fd;
 		free(target);
 	}
 	return true;
@@ -1053,11 +1051,20 @@ static int set_pidfile(char *pidfile)
 	return fd;
 }
 
+static void close_fd_hierarchies(void)
+{
+	int i;
+	for (i = 0; i < num_hierarchies; i++)
+		if (fd_hierarchies[i] > 0)
+			close(fd_hierarchies[i]);
+}
+
 int main(int argc, char *argv[])
 {
-	int ret = -1, pidfd;
+	int ret = EXIT_FAILURE;
+	int pidfd = -1;
 	char *pidfile = NULL, *v = NULL;
-	size_t pidfile_len;
+	size_t pidfile_len, mmap_len = 0;
 	/*
 	 * what we pass to fuse_main is:
 	 * argv[0] -s -f -o allow_other,directio argv[1] NULL
@@ -1071,7 +1078,7 @@ int main(int argc, char *argv[])
 	if (swallow_option(&argc, argv, "-o", &v)) {
 		if (strcmp(v, "allow_other") != 0) {
 			fprintf(stderr, "Warning: unexpected fuse option %s\n", v);
-			exit(1);
+			exit(EXIT_FAILURE);
 		}
 		free(v);
 		v = NULL;
@@ -1081,7 +1088,7 @@ int main(int argc, char *argv[])
 
 	if (argc == 2  && strcmp(argv[1], "--version") == 0) {
 		fprintf(stderr, "%s\n", VERSION);
-		exit(0);
+		exit(EXIT_SUCCESS);
 	}
 	if (argc != 2 || is_help(argv[1]))
 		usage(argv[0]);
@@ -1089,7 +1096,7 @@ int main(int argc, char *argv[])
 	do_reload();
 	if (signal(SIGUSR1, reload_handler) == SIG_ERR) {
 		fprintf(stderr, "Error setting USR1 signal handler: %m\n");
-		exit(1);
+		goto out;
 	}
 
 	newargv[cnt++] = argv[0];
@@ -1100,7 +1107,8 @@ int main(int argc, char *argv[])
 	newargv[cnt++] = NULL;
 
 	/* Share memory with our clone(). */
-	fd_hierarchies = mmap(NULL, sizeof(int *) * num_hierarchies, PROT_READ | PROT_WRITE, MAP_SHARED | MAP_ANONYMOUS, -1, 0);
+	mmap_len = sizeof(int *) * num_hierarchies;
+	fd_hierarchies = mmap(NULL, mmap_len, PROT_READ | PROT_WRITE, MAP_SHARED | MAP_ANONYMOUS, -1, 0);
 	if (!fd_hierarchies)
 		goto out;
 
@@ -1124,12 +1132,18 @@ int main(int argc, char *argv[])
 	if ((pidfd = set_pidfile(pidfile)) < 0)
 		goto out;
 
-	ret = fuse_main(nargs, newargv, &lxcfs_ops, NULL);
-
-	dlclose(dlopen_handle);
-	unlink(pidfile);
-	close(pidfd);
+	if (!fuse_main(nargs, newargv, &lxcfs_ops, NULL))
+		ret = EXIT_SUCCESS;
 
 out:
-	return ret;
+	if (dlopen_handle)
+		dlclose(dlopen_handle);
+	if (pidfile)
+		unlink(pidfile);
+	if (pidfd > 0)
+		close(pidfd);
+	close_fd_hierarchies();
+	if (fd_hierarchies)
+		munmap(fd_hierarchies, mmap_len);
+	exit(ret);
 }
--- a/lxcfs.c
+++ b/lxcfs.c
@@ -27,9 +27,25 @@
 #include <sys/epoll.h>
 #include <sys/mount.h>
 #include <sys/socket.h>
+#include <sys/syscall.h>
 
 #include "bindings.h"
 #include "config.h" // for VERSION
+
+/* Define pivot_root() if missing from the C library */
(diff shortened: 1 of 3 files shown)


brauner:
Should fix https://github.com/lxc/lxcfs/issues/123 and https://github.com/docker/docker/issues/24834.

brauner:
This is probably not the final version. @hallyn, I'd like to hear your impression/ideas/improvements. :)

lxcfs.c
```diff
@@ -82,22 +186,27 @@ static void do_reload(void)
 		goto good;
 
 	dlopen_handle = dlopen("/usr/lib/lxcfs/liblxcfs.so", RTLD_LAZY);
```
brauner: Maybe we want `RTLD_NOW` here?

hallyn: Yes, did you see no errors with lazy?  That's surprising, though I suppose there's reasonable likelyhood that everything we need was already needed during startup, bc we do a lot of stuff during startup.

hallyn: Initial impression - looks great, thanks much.  I had a few comments last night, will post them later.  (It's possible I'll answer my own questions on a second read-through).

brauner: I'm happy to simplify this wherever possible. Maybe we can simplify keeping track of the namespace we are currently in aka avoiding the use of a struct for this.

lxcfs.c
```diff
@@ -797,6 +813,133 @@ static bool umount_if_mounted(void)
...
+ *	- /var/lib/lxcfs		<-- the fuse mount
+ *	- /dev				<-- because of /dev/fuse
+ * 	- /sys				<-- because we want to mount /sys/fs/connections/fuse
+ * 	- /sys/fs/connections/fuse	<-- because of fuse
+ * 	- /proc				<-- where we read info from
+ * (Is that all we need? Did we not pin any unnecessary mounts?)
+ */
+static int pivot_prepare(void)
+{
+	if (mkdir(ROOTDIR, 0755) < 0 && errno != EEXIST) {
```
hallyn: Would 0700 work here?

lxcfs.c
```diff
@@ -83,12 +99,84 @@ static void users_unlock(void)
...
+ * Functions and types used to reload dynamic library.
+ */
 static volatile sig_atomic_t need_reload;
 
 /* do_reload - reload the dynamic library.  Done under
  * lock and when we know the user_count was 0 */
-static void do_reload(void)
+static void do_reload(struct preserved_ns *ns)
 {
+	if (ns->active_ns != -1) {
```
hallyn: I don't understand this check for active_ns == -1.  It seems to me it should never be -1 unless preserve_ns() failed?  Relatedly, preserve_ns() takes the 'is_caller_pid' bool argument, but it is always passed in as true.

hallyn: Also the struct preserved_ns will work as is so no need to change it, but I'm kind of wondering why you didn't just keep two global ints for the two fds.  This doesn't need to generalize right?  I'm probably missing something, hence I'm asking...

lxcfs.c
brauner: Sure, I totally forgot that we use 0700.

lxcfs.c
brauner: So the idea was to allow for some generality in case we need to expand this for some reason. But we might simply go for a more simple solution. So we could use two global ints for the two namespace and another global int for the current active namespace.(?)

lxcfs.c
```diff
@@ -789,25 +898,152 @@ static bool mkdir_p(const char *dir, mode_t mode)
...
+		close(oldroot);
+	if (newroot != -1)
+		close(newroot);
+	return -1;
+}
+
+/*
+ * Prepare our new root: We need to mount everything that fuse needs to
+ * correctly work in our minimal chroot:
+ *	- /var/lib/lxcfs		<-- the fuse mount
```
brauner: Currently, we're not mounting `/var/lib/lxcfs` in our private mount namespace. I wonder whether we need to. I'll leave it unmounted for now.

lxcfs.c
hallyn: Quoting Christian Brauner (notifications@github.com):
> >  {
> > +	if (ns->active_ns != -1) {
> 
> So the idea was to allow for some generality in case we need to expand this for some reason. But we might simply go for a more simple solution. So we could use two global ints for the two namespace and another global int for the current active namespace.(?)

Practically speaking, the biggest concern is to avoid having code paths which
aren't used/tested at all, which we might accidentally fall into at some point.

brauner: Yeah, I already started simplifying it. I may push it today or tomorrow, depending on when I get to work on it.

lxcfs.c
```diff
@@ -67,30 +83,107 @@ static void users_unlock(void)
...
+/*
+ * Functions and types used to reload dynamic library.
+ */
 static volatile sig_atomic_t need_reload;
 
 /* do_reload - reload the dynamic library.  Done under
  * lock and when we know the user_count was 0 */
 static void do_reload(void)
 {
+	if ((mnt_ns_active != -1) && (mnt_ns_active != mnt_ns_init))
```
brauner: I think that this is a valid case since `do_reload()` is called in `main()` before we have called `preserve_ns()`.

lxcfs.c
```diff
@@ -67,30 +83,107 @@ static void users_unlock(void)
...
-		_exit(1);
-	}
+	dlopen_handle = dlopen("/usr/lib/lxcfs/liblxcfs.so", RTLD_NOW);
+	if (!dlopen_handle)
+		goto bad;
 
 good:
 	if (need_reload)
 		fprintf(stderr, "lxcfs: reloaded\n");
+	if (mnt_ns_active != -1 && (mnt_ns_active != mnt_ns_lxcfs))
```
brauner: See https://github.com/lxc/lxcfs/pull/124/files#r72824002.

brauner:
Changelog 2016-07-29:
- lxcfs: fix logic in `main()`
  - perform all necessary cleanup operations on failure
  - use `EXIT_*` macros

- lxcfs: simplify storing and switch namespaces
  - replace the struct with three simple global integers
  - simplify namespace handling functions

- lxcfs: more restrictive perms for `ROOTDIR` (`0700`)
- Before we switch to our private mount namespace and pivot root we need to make sure that all symbols we need are properly resolved. So use `RTLD_NOW` for dynamic library reload

brauner:
Changelog 2016-07-30:
- simplify exit logic in `pivot_enter()`

hallyn:
I did the following:

built a lxcfs package based on your branch.  installed the package.  Rebooted.  started a ubuntu xenial lxd container.  Ran 'uptime', it worked.  Exited the container, re-installed the package.  re-attached to the container, uptime failed:

root@x1:~# cat /proc/uptime
cat: /proc/uptime: Transport endpoint is not connected

So re-exec somehow is failing.

brauner:
Hm, I'm not on a computer right now. I'll look into this tomorrow or later tonight.

brauner:
I currently can't reproduce this. The only circumstance where this happens is when I try to restart the service via `systemctl restart lxcfs` nultiple times and systemd errors complaining that I restart too often.

brauner:
But I suspect that this is caused by `fuse_umount` not being executed.

brauner:
As we've discussed on irc this is `setns()` telling us that we're trying to reassociate the caller with a new mount namespace while we're multi-threaded. I'm currently weighing options.

brauner:
Ok, my first idea was to somehow use `pthread_atfork()` handlers and `fork()` a new child to guarantee that we are single-threaded. But I find that to be very undesirable and I doubt that `pthread_atfork()` and `dlopen()` get along very well. Other options of becoming temporarily single-threaded before switching namespaces especially in combinations with `fuse_loop_mt()` are not known to me. So we have two options, I think:

1. Keep everything as is (aka `pivot_root()` + `umount2(oldroot, MNT_DETACH)`) and call `do_reload()` in the private lxcfs namespace presupposing that everything we need to successfully reload the dynamic library with `dlopen()` is in fact present.
2. `unshare(CLONE_NEWNS)` + remount `/` `MS_REC | MS_SLAVE` to propagate mount events from our master into the slave mount, and then still `pivot_root()`. I don't know how likely it is that we pin any mounts when `/` in the initial namespace is mounted `MS_PRIVATE` (which shouldn't be the case with nearly all init systems. At least `systemd` guarantees that `/` is mounted `MS_SHARED` to satisfy container runtimes.

Further ideas or preferences? So far I've implemented 2.

hallyn:
Can we do something like:

at startup, fork a task into a new mntns.  Have it mount
/var/run/lxcfs/controllers MS_PRIVATE and do the mounts under that.
Have it open one fd per mounted /var/run/lxcfs/controllers/$controller
directory, and pass them back to the main lxcfs task.  That task then
uses openat() with those fds instead instead of
open(/var/run/lxcfs/controllers/$controller).

brauner:
On Mon, Aug 01, 2016 at 08:50:29PM -0700, Serge Hallyn wrote:
> Can we do something like:
> 
> at startup, fork a task into a new mntns.  Have it mount
> /var/run/lxcfs/controllers MS_PRIVATE and do the mounts under that.
> Have it open one fd per mounted /var/run/lxcfs/controllers/$controller
> directory, and pass them back to the main lxcfs task.  That task then

So I've written a small test program and yeah, we can use
clone(..., CLONE_NEWNS), use a socketpair() to send file descriptors from
the lxcfs mntns to the initial mnts and use openat() to get access to files.

> uses openat() with those fds instead instead of
> open(/var/run/lxcfs/controllers/$controller).

I need a more thorough understanding of how FUSE works and how to modify
lxcfs_opendir()/cg_opendir() and lxcfs_open()/cg_open() (I guess that's what we
want to change.) and where openat() is supposed to go.

brauner:
On Tue, Aug 02, 2016 at 10:17:51PM +0200, Christian Brauner wrote:
> On Mon, Aug 01, 2016 at 08:50:29PM -0700, Serge Hallyn wrote:
> > Can we do something like:
> > 
> > at startup, fork a task into a new mntns.  Have it mount
> > /var/run/lxcfs/controllers MS_PRIVATE and do the mounts under that.
> > Have it open one fd per mounted /var/run/lxcfs/controllers/$controller
> > directory, and pass them back to the main lxcfs task.  That task then
> 
> So I've written a small test program and yeah, we can use
> clone(..., CLONE_NEWNS), use a socketpair() to send file descriptors from
> the lxcfs mntns to the initial mnts and use openat() to get access to files.
> 
> > uses openat() with those fds instead instead of
> > open(/var/run/lxcfs/controllers/$controller).
> 
> I need a more thorough understanding of how FUSE works and how to modify
> lxcfs_opendir()/cg_opendir() and lxcfs_open()/cg_open() (I guess that's what we
> want to change.) and where openat() is supposed to go.

So let's keep a global static struct that serves as an open file descriptor
table for the private cgroup mounts and which gets filled on lxcfs startup and
is considered read-only after that. This open file descriptor table can then be
used by the callbacks defined for fuse_operations to lookup paths. That *should*
work.

hallyn:
Sounds good, thanks.

brauner:
[PATCH v2] move lxcfs to minimal chroot

This is a new approach. Parts of which have been discussed above but the actual design diverges. After the constructor has run and collected the mounted hierarchies we `clone()` a new process which mounts the private lxcfs cgroup mounts in a new private mount namespace not accessible to other processes. We then open an fd for each mounted hierarchy and use `CLONE_FILES` to keep them valid in both the child and the parent. This allows us to avoid using `SCM` credentials. Before the clone we `mmap()` and array of fds of equal size to the number of mounted controllers. This fd array will be used by the child to store the opened fds referring to the private lxcfs cgroup mounts in the new mount namespace. The mounted hierarchies array in conjunction with the opened file descriptor array will then serve as an open file descriptor table for the fuse callbacks in `bindings.c`. To make this possible we make heavy use of the `*at()` family of functions. To have fuse perform all operations on the private lxcfs cgroup mounts we need to ensure that only relative paths are passed to `*at()`-style functions.

lxcfs.c
```diff
@@ -1053,11 +1051,20 @@ static int set_pidfile(char *pidfile)
 	return fd;
 }
 
+static void close_fd_hierarchies(void)
+{
+	int i;
+	for (i = 0; i < num_hierarchies; i++)
+		if (fd_hierarchies[i] > 0)
```
hallyn: nit - >= 0.

brauner:
If you could build a package and do a quick test that would be great. My system is slightly busted right now.

brauner:
I need to move the initialization of `fd_hierarchies` into the constructor. Otherwise we'll run into trouble.

brauner:
Ok, I built a package locally, installed and pounded on `lxcfs` with a few containers and also tried reloading the shared library with:

```
lxc-start -n c
sudo kill -s SIGUSR1 LXCFS_PID
lxc-stop -n c
lxc-start -n c
```

that all seems to work.

brauner:
Changelog 2016-08-06:
  - initialize `fd_hierarchies` during `__constructor__` run
  - mark `fd_hierarchies`, `hierarchies`, and `num_hierarchies` as read-only
  - when closing `fd_hierarchies[i]` check `if (hierarchies[i] >= 0)` and not just `if (hierarchies[i] > 0)`
  - order includes in `bindings.c`
  - only mount `/run/lxcfs/*` in private mount namespace, we do not need `/proc`, `/dev`, `/sys`, and `/sys/fs/fuse/connections` anymore.

hallyn:
After re-installing the .deb a running container's lxcfs-provided procfiles continue to work, so this is looking good.  Is it ready for merging in your view?

hallyn:
Maybe it's too late at night but you seem to be filling in the fd_hierarchies() before you pivot_root.  Once you have the open fds you don't need the task that open()ed them at all any more do you?

brauner:
I think it's ready to merge. About the pivot_root part: you mean we should move the pivot root out of the clone task and possibly also the constructor?

hallyn:
Quoting Christian Brauner (notifications@github.com):
> I think it's ready to merge. About the pivot_root part: you mean we should move the pivot root out of the clone task and possibly also the constructor?

Right.  The cloned task was only needed to open the fds, right?  And it
doesn't need to be pivot_root()ed, it only needed to be in a private mntns
so it could mount the cgfs temporarily privately to open the fds.
I suppose technically it could have been done without another task, just
doing unshare() and then setns()ing back to the original ns at the end.
We don't need the new ns to be kept open, it'll automatically stick around
so long as we have the fds open.

brauner:
I can code this up when I get home...

brauner:
On Mon, Aug 08, 2016 at 07:57:54AM -0700, Serge Hallyn wrote:
> Quoting Christian Brauner (notifications@github.com):
> > I think it's ready to merge. About the pivot_root part: you mean we should move the pivot root out of the clone task and possibly also the constructor?
> 
> Right.  The cloned task was only needed to open the fds, right?  And it
> doesn't need to be pivot_root()ed, it only needed to be in a private mntns
> so it could mount the cgfs temporarily privately to open the fds.
> I suppose technically it could have been done without another task, just
> doing unshare() and then setns()ing back to the original ns at the end.
> We don't need the new ns to be kept open, it'll automatically stick around
> so long as we have the fds open.

My original concern came from cgmanager. It used to pin mounts (I think
autofs.). When it moved in to a private namespace it didn't umount2()
and a process outside e.g. couldn't move the mount. So what we should do is:

	- preserve initial mount namespace
	- unshare(CLONE_NEWNS)
	- remount root (MS_REC | MS_PRIVATE)
	- umount2("/", MS_DETACH)
	- mount all cgroup hierarchies on /run/lxcfs/cgroup/$CONTROLLERS
	- open file descriptors for all mounted hierarchies
	- setns() to initial mount namespace

So your point is only about avoiding pivot root? Or are you also saying that
umount2("/", MS_DETACH) is not necessary? That would surprise me since I thought
that the lifetime of a mount (unless unmounted) is the lifetime of its
corresponding mount namespace. And since we keep the mount namespace alive for
the lifetime of the shared library a missing umount2("/", MS_DETACH) will cause
us to pin mounts we don't want to.

brauner:
Actually I'm not even sure whether you can avoid to pivot root when you want to
`umount2("/", MS_DETACH)`.

hallyn:
Hm, yeah, I suppose we do need it.  Sounds good.

brauner:
Changelog 2016-08-08:
  - remove `lxcfs_clone()`. Instead of `clone(..., CLONE_NEWNS | CLONE_FILES)` simply use `unshare(CLONE_NEWNS`.
  - add `preserve_ns()` to preserve initial mount namespace on library start up before we `unshare(CLONE_NEWNS)`.
  - change logic: `preserve_ns(getpid())` -->  `unshare(CLONE_NEWNS)` -->  perform the private cgroup mounts for lxcfs --> pivot root and `umount2("/", MNT_DETACH)` --> `setns(fd /* initial mount namespace */)`

PR #127, opened 2016-08-21
2016 08 21/bugfixes

### LINKED ISSUE #116: Allow chdir to the paths the user can otherwise access
This is from: https://bugs.launchpad.net/ubuntu/+source/lxcfs/+bug/1587688

Basically, we currently have this weird behavior on cd:
 - /var/lib/lxcfs => EINVAL
 - /var/lib/lxcfs/proc => EACCESS
 - /var/lib/lxcfs/cgroup => EIO

All of those should succeed so long as the user otherwise has permissions to view those paths (which should always be the case for those three).

It would probably also be nice to make sure that all paths under /var/lib/lxcfs probably return a "." and ".." entry and plausible ownership and permissions for all of those.

That would make lxcfs behave a bit more like an actual filesystem.

This is a nice to have, probably beginner contributor kind of issue as it's not at all required for the normal operation of lxcfs.

brauner:
closes #116.

### DIFF
--- a/bindings.c
+++ b/bindings.c
@@ -1605,7 +1605,7 @@ int cg_getattr(const char *path, struct stat *sb)
 
 	controller = pick_controller_from_path(fc, path);
 	if (!controller)
-		return -EIO;
+		return -errno;
 	cgroup = find_cgroup_in_path(path);
 	if (!cgroup) {
 		/* this is just /cgroup/controller, return it as a dir */
@@ -1705,7 +1705,7 @@ int cg_opendir(const char *path, struct fuse_file_info *fi)
 		// return list of keys for the controller, and list of child cgroups
 		controller = pick_controller_from_path(fc, path);
 		if (!controller)
-			return -EIO;
+			return -errno;
 
 		cgroup = find_cgroup_in_path(path);
 		if (!cgroup) {
@@ -1863,7 +1863,7 @@ int cg_open(const char *path, struct fuse_file_info *fi)
 
 	controller = pick_controller_from_path(fc, path);
 	if (!controller)
-		return -EIO;
+		return -errno;
 	cgroup = find_cgroup_in_path(path);
 	if (!cgroup)
 		return -EINVAL;
@@ -1937,7 +1937,7 @@ int cg_access(const char *path, int mode)
 
 	controller = pick_controller_from_path(fc, path);
 	if (!controller)
-		return -EIO;
+		return -errno;
 	cgroup = find_cgroup_in_path(path);
 	if (!cgroup) {
 		// access("/sys/fs/cgroup/systemd", mode) - rx allowed, w not
@@ -2724,7 +2724,7 @@ int cg_chown(const char *path, uid_t uid, gid_t gid)
 
 	controller = pick_controller_from_path(fc, path);
 	if (!controller)
-		return -EINVAL;
+		return -errno;
 	cgroup = find_cgroup_in_path(path);
 	if (!cgroup)
 		/* this is just /cgroup/controller */
@@ -2789,7 +2789,7 @@ int cg_chmod(const char *path, mode_t mode)
 
 	controller = pick_controller_from_path(fc, path);
 	if (!controller)
-		return -EINVAL;
+		return -errno;
 	cgroup = find_cgroup_in_path(path);
 	if (!cgroup)
 		/* this is just /cgroup/controller */
@@ -2854,7 +2854,7 @@ int cg_mkdir(const char *path, mode_t mode)
 
 	controller = pick_controller_from_path(fc, path);
 	if (!controller)
-		return -EINVAL;
+		return errno == ENOENT ? -EPERM : -errno;
 
 	cgroup = find_cgroup_in_path(path);
 	if (!cgroup)
@@ -2875,7 +2875,7 @@ int cg_mkdir(const char *path, mode_t mode)
 		else if (last && strcmp(next, last) == 0)
 			ret = -EEXIST;
 		else
-			ret = -ENOENT;
+			ret = -EPERM;
 		goto out;
 	}
 
@@ -2908,7 +2908,7 @@ int cg_rmdir(const char *path)
 
 	controller = pick_controller_from_path(fc, path);
 	if (!controller)
-		return -EINVAL;
+		return -errno;
 
 	cgroup = find_cgroup_in_path(path);
 	if (!cgroup)
--- a/bindings.c
+++ b/bindings.c
@@ -1508,9 +1508,9 @@ static char *pick_controller_from_path(struct fuse_context *fc, const char *path
 
 	if (strlen(path) < 9)
 		return NULL;
-	if (*(path+7) != '/')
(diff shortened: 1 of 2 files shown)


bindings.c
```diff
@@ -4065,8 +4076,8 @@ int proc_open(const char *path, struct fuse_file_info *fi)
 int proc_access(const char *path, int mask)
 {
 	/* these are all read-only */
-	if ((mask & ~R_OK) != 0)
-		return -EACCES;
+	if (access(path, mask) < 0)
```
brauner: Not sure, if we should just check for `R_OK`.

brauner:
#boredontheAutobahn

bindings.c
hallyn: Can you explain why you'd want this to pass if the user asked about W_OK?

bindings.c
```diff
@@ -1909,11 +1909,17 @@ int cg_open(const char *path, struct fuse_file_info *fi)
...
 	const char *cgroup;
-	char *last = NULL, *path1, *path2, * cgdir = NULL, *controller;
+	char *path1, *path2, *controller;
+	char *last = NULL, *cgdir = NULL;
 	struct cgfs_files *k = NULL;
 	struct fuse_context *fc = fuse_get_context();
-	int ret;
+
+	if (strcmp(path, "/cgroup") == 0) {
+		ret = 0;
```
hallyn: I think you should check the mode for write access and return an error.  Is there a reason not to?

hallyn:
Shouldn't you be *driving*? :-)

hallyn:
LGTM apart from those the comments I posted, thanks.

bindings.c
brauner: Right, totally missed the read-only line above. Users should still be able to access `/var/lib/lxcfs/proc`. I think I'll just do:

```
if (strcmp(path, "/proc") == 0)
        return 0;
```

and leave the original check unchanged.

bindings.c
brauner: `/sys/fs/cgroup` has perms `755` . I think users should be able to acces `/var/lib/lxcfs/cgroup` in the same way.

brauner: `cg_getattr()` will currently also prevent any other file or directory to be created since for obvious reasons `pick_controller_from_path()` will return `NULL`. We can merge this and I'll make the handling of this case smarter later.

brauner:
As promised, some more changes (https://github.com/lxc/lxcfs/pull/127/commits/d71858ab17b536535491e7ae9447bf76c25354e9, https://github.com/lxc/lxcfs/pull/127/commits/53d00b24d3d904687f4dfad812b1b95d5b151d9d, https://github.com/lxc/lxcfs/pull/127/commits/fe1a0797d2c97c82aa9456588c4c50f80692c91e) to be more consistent. (This time not straight from the Autobahn. :))

lxcfs.c
```diff
@@ -489,6 +497,10 @@ static int lxcfs_readdir(const char *path, void *buf, fuse_fill_dir_t filler, of
 static int lxcfs_access(const char *path, int mode)
 {
 	int ret;
+
+	if (strcmp(path, "/") == 0 && access(path, R_OK) == 0)
```
hallyn: I don't think we need this - the admin is not meant to change perms on /var/lib/lxcfs itself to deny read access.  what I meant was checking that the mode that is passed in does not contain write.

brauner:
Changelog 2016-08-22:
  - Check that the mode passed in to `lxcfs_access()` does not contain write.

lxcfs.c
```diff
@@ -493,14 +493,14 @@ static int lxcfs_readdir(const char *path, void *buf, fuse_fill_dir_t filler, of
...
+	return -ENOENT;
 }
 
 static int lxcfs_access(const char *path, int mode)
 {
 	int ret;
 
-	if (strcmp(path, "/") == 0 && access(path, R_OK) == 0)
+	if (strcmp(path, "/") == 0 && (mode & W_OK) == 0)
 		return 0;
```
hallyn: -EACCESS?

bindings.c
```diff
@@ -1907,11 +1907,18 @@ int cg_open(const char *path, struct fuse_file_info *fi)
...
 	const char *cgroup;
-	char *last = NULL, *path1, *path2, * cgdir = NULL, *controller;
+	char *path1, *path2, *controller;
+	char *last = NULL, *cgdir = NULL;
 	struct cgfs_files *k = NULL;
 	struct fuse_context *fc = fuse_get_context();
-	int ret;
+
+	if (strcmp(path, "/cgroup") == 0) {
+		if (access(path, W_OK) == 0)
```
hallyn: instead of access(path, W_OK) did you mean if (mode & W_OK) ?

lxcfs.c
```diff
@@ -478,7 +478,7 @@ static int lxcfs_readdir(const char *path, void *buf, fuse_fill_dir_t filler, of
 		    filler(buf, "..", NULL, 0) != 0 ||
 		    filler(buf, "proc", NULL, 0) != 0 ||
 		    filler(buf, "cgroup", NULL, 0) != 0)
-			return -EINVAL;
+			return -ENOENT;
```
hallyn: So looking through other fuse examples, everyone else seems to simply ignore filler() errors...  which I don't think is the right thing to do, but ENOENT just seems the wrong value.  Oh actually fill_dir_old() in lib/fuse.c itself returns -ENOMEM if filler fails - that seems best.

hallyn:
Ok I think I'm done nitpicking once those are addressed - thanks.  (Please feel free to ping when updated, or just merge yourself)

brauner:
@hallyn can't merge here. I have no write access currently. :)

bindings.c
```diff
@@ -1909,18 +1919,25 @@ int cg_open(const char *path, struct fuse_file_info *fi)
...
 	const char *cgroup;
-	char *last = NULL, *path1, *path2, * cgdir = NULL, *controller;
+	char *path1, *path2, *controller;
+	char *last = NULL, *cgdir = NULL;
 	struct cgfs_files *k = NULL;
 	struct fuse_context *fc = fuse_get_context();
-	int ret;
+
+	if (strcmp(path, "/cgroup") == 0) {
+		if ((mode & W_OK) == 0)
```
brauner: Actually, I'm not sure if this is really needed. Even if the cgroup root mount is writeable operations like `cg_mkdir()` et al. will fail with `EPERM`. Basically *all* operations will fail on the cgroup root mount point because they try to pick a controller from this path and if they do not find one they return an error. For example, `cg_mkdir()` does:

```
controller = pick_controller_from_path(fc, path);
if (!controller)
        return errno == ENOENT ? -EPERM : -errno;
```


My idea was that users should still be able to descend into `/var/lib/lxcfs/cgroup` but not be able to e.g. do `touch` or `mkdir`. So what do you think about removing the check here?

brauner: Yeah, I think I'll remove the `mode & W_OK` check for the reason outlined above.

PR #129, opened 2016-08-22
bindings: use openat fd for fstatat(), unlinkat()

brauner:

### DIFF
--- a/bindings.c
+++ b/bindings.c
@@ -571,7 +571,7 @@ int cgfs_create(const char *controller, const char *cg, uid_t uid, gid_t gid)
 	return 0;
 }
 
-static bool recursive_rmdir(const char *dirname, int fd, int cfd)
+static bool recursive_rmdir(const char *dirname, int fd, const int cfd)
 {
 	struct dirent *direntp;
 	DIR *dir;
@@ -586,6 +586,7 @@ static bool recursive_rmdir(const char *dirname, int fd, int cfd)
 	dir = fdopendir(dupfd);
 	if (!dir) {
 		lxcfs_debug("Failed to open %s: %s.\n", dirname, strerror(errno));
+		close(dupfd);
 		return false;
 	}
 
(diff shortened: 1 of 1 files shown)


brauner:
Otherwise `fstatat()` will fail and commands like `rmdir()` will not work.

bindings.c
```diff
@@ -654,7 +651,7 @@ bool cgfs_remove(const char *controller, const char *cg)
 	if (fd < 0)
```
hallyn: So is there any reason to still open fd here?

hallyn: Oh, I guess so.

Could you also make sure that when recursive_rmdir() fails, fd gets closed, and when the fdopendir there fails, dupfd gets closed?

brauner: Done.

brauner:
Are you fine with this now? :)

hallyn:
Hi,

having fd be closed as a side effect of recursive_rmdir() just takes a bit too much to think about.  And I think it's leaking right now if the dup() fails, and being closed prematurely for nested calls.

Let's just have recursive_rmdir() accept the open fd and not close it, and have cgfs_remove() do

    bret = recursive_rmdir(dirname, fd);
    close(fd);
    return bret;

brauner:
`fd` is not closed as a side-effect only the `dup()`ed is, no? The problem is that `fstatat()` and `unlinkat()` need the original `fd` on which `openat()` is called.

bindings.c
```diff
@@ -652,7 +652,9 @@ bool cgfs_remove(const char *controller, const char *cg)
...
 		return false;
 
-	return recursive_rmdir(dirnam, fd);
+	bret = recursive_rmdir(dirnam, fd, cfd);
```
brauner: Here `cfd` needs to be passed down into `recursive_rmdir()` because `unlinkat()` and `fstatat()` need. `fd` itself is opened in the lines above via `openat(cfd, ...)` and therefore useless for `fstatat()` and `unlinkat()` if I'm not mistaken.

hallyn: Sorry I got mixed up between the patch and the unpatched file.

PR #132, opened 2016-08-24
bindings: improve debugging

brauner:
- replace multiple DEBUG ifdefines with a single ifdefine at the top
- ifdefine lxcfs_debug() macro function that expands to nothing when -DDEBUG is
  not given


hallyn:
Thanks!

PR #136, opened 2016-08-26
bindings: revert cgroup check and improve fs behavior

brauner:
We do not need to check whether mode & W_OK is passed in. Even if the cgroup
root mount is writeable operations like cg_mkdir() et al. will fail with e.g.
EPERM. Basically all operations will fail on the cgroup root mount point because
the first operation they perform is pick_controller_from_path(). That is to say
they try to e.g. pick "blkio" from /var/lib/lxcfs/cgroup/blkio/some/cgroups an
similiar for all other controllers. If pick_controller_from_path() fails they
all return an appropriate errno. For example, cg_mkdir() does:

        controller = pick_controller_from_path(fc, path);
        if (!controller)
                return errno == ENOENT ? -EPERM : -errno;

This means, we do not need to return an errno already in cg_access when
mode & W_OK is passed in. This has the advantage that users are still able to
descend into /var/lib/lxcfs/cgroup via:

        cd /var/lib/lxcfs/cgroup

but are still blocked from doing any write operations.

### DIFF
--- a/bindings.c
+++ b/bindings.c
@@ -1926,11 +1926,8 @@ int cg_access(const char *path, int mode)
 	struct cgfs_files *k = NULL;
 	struct fuse_context *fc = fuse_get_context();
 
-	if (strcmp(path, "/cgroup") == 0) {
-		if ((mode & W_OK) == 0)
-			return -EACCES;
+	if (strcmp(path, "/cgroup") == 0)
 		return 0;
-	}
 
 	if (!fc)
 		return -EIO;
--- a/bindings.c
+++ b/bindings.c
@@ -1541,12 +1541,17 @@ static const char *find_cgroup_in_path(const char *path)
 {
 	const char *p1;
 
-	if (strlen(path) < 9)
+	if (strlen(path) < 9) {
+		errno = EINVAL;
 		return NULL;
-	p1 = strstr(path+8, "/");
-	if (!p1)
+	}
+	p1 = strstr(path + 8, "/");
+	if (!p1) {
+		errno = EINVAL;
 		return NULL;
-	return p1+1;
+	}
+	errno = 0;
+	return p1 + 1;
 }
 
 /*
@@ -1866,7 +1871,7 @@ int cg_open(const char *path, struct fuse_file_info *fi)
 		return -errno;
 	cgroup = find_cgroup_in_path(path);
 	if (!cgroup)
-		return -EINVAL;
+		return -errno;
 
 	get_cgdir_and_path(cgroup, &cgdir, &last);
 	if (!last) {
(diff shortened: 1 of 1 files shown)


brauner:
```
root@conventiont:~# cd /var/lib/lxcfs/cgroup/
root@conventiont:/var/lib/lxcfs/cgroup# ls -a
.  ..  blkio  cpu,cpuacct  cpuset  devices  freezer  hugetlb  memory  name=systemd  net_cls,net_prio  perf_event  pids

root@conventiont:/var/lib/lxcfs/cgroup# mkdir test
mkdir: cannot create directory ‘test’: Operation not permitted

root@conventiont:/var/lib/lxcfs/cgroup# chown 0:0 blkio
chown: changing ownership of 'blkio': Invalid argument

root@conventiont:/var/lib/lxcfs/cgroup# chmod 600 blkio
chmod: changing permissions of 'blkio': Invalid argument

root@conventiont:/var/lib/lxcfs/cgroup# chmod 755 blkio
chmod: changing permissions of 'blkio': Invalid argument

root@conventiont:/var/lib/lxcfs/cgroup# touch aaa
touch: cannot touch 'aaa': Function not implemented
```

bindings.c
```diff
@@ -2905,11 +2910,11 @@ int cg_rmdir(const char *path)
 
 	controller = pick_controller_from_path(fc, path);
 	if (!controller)
-		return -errno;
+		return errno == ENOENT ? -EPERM : -errno;
```
hallyn: Why -EPERM if it does not exist?

bindings.c
```diff
@@ -2848,14 +2854,13 @@ int cg_mkdir(const char *path, mode_t mode)
...
 
-
 	controller = pick_controller_from_path(fc, path);
 	if (!controller)
 		return errno == ENOENT ? -EPERM : -errno;
 
 	cgroup = find_cgroup_in_path(path);
 	if (!cgroup)
-		return -EINVAL;
+		return -errno;
```
hallyn: find_cgroup_in_path does not set errno yet though right?

brauner: It sets an `errno` so it's safe to return an `errno` directly here.

bindings.c
brauner: Good point, missed that one.

brauner: So you referred to `cg_rmdir()` here? I adapted that already in the updated PR but somehow it is still shown here. Not sure why. Just look at the patch https://github.com/lxc/lxcfs/pull/136.patch :)

brauner: I actually tend to disagree with my former consent to your implicit `-EPERM` objection. :)
We should be able to safely return -EPERM for:

```
 	if (!pick_controller_from_path())
                /* Someone's trying to delete "/cgroup". */

 	if (!find_cgroup_in_path())
 	        /* Someone's trying to delete a controller e.g. "/blkio". */

 	if (!get_cgdir_and_path()) {
 		/* Someone's trying to delete a cgroup on the same level as the
 		 * "/lxc" cgroup e.g. rmdir "/cgroup/blkio/lxc" or
 		 * rmdir "/cgroup/blkio/init.slice".
 		 */
 	}
```
All other interesting cases are caught further down.

PR #137, opened 2016-08-27
2016 08 27/postmodernism branch

### LINKED ISSUE #135: errors when trying `lxcfs -h` as a user
Ohai,

```
% ./lxcfs -h
failed to create directory '/run/lxcfs/': Permission deniedFailed to create lxcfs cgroup mountpoint.
Usage:

lxcfs [-p pidfile] mountpoint
  Default pidfile is /run/lxcfs.pid
lxcfs -h
```

Why does lxcfs try to do something while I only wanted to see the help? This is esp annoying as this line is added to the manpage if the build is not running as root :)

brauner:
My first postmodern commit. Closes #135.

### DIFF
--- a/bindings.c
+++ b/bindings.c
@@ -32,16 +32,6 @@
 #include "bindings.h"
 #include "config.h" // for VERSION
 
-#ifdef DEBUG
-#define lxcfs_debug(format, ...)                                               \
-	do {                                                                   \
-		fprintf(stderr, "%s: %d: %s: " format, __FILE__, __LINE__,     \
-			__func__, __VA_ARGS__);                                \
-	} while (false)
-#else
-#define lxcfs_debug(format, ...)
-#endif /* DEBUG */
-
 enum {
 	LXC_TYPE_CGDIR,
 	LXC_TYPE_CGFILE,
--- a/bindings.c
+++ b/bindings.c
@@ -28,25 +28,9 @@
 #include <sys/mount.h>
 #include <sys/param.h>
 #include <sys/socket.h>
-#include <sys/syscall.h>
 
 #include "bindings.h"
 #include "config.h" // for VERSION
-
-/* Define pivot_root() if missing from the C library */
-#ifndef HAVE_PIVOT_ROOT
-static int pivot_root(const char * new_root, const char * put_old)
-{
-#ifdef __NR_pivot_root
-return syscall(__NR_pivot_root, new_root, put_old);
-#else
-errno = ENOSYS;
-return -1;
-#endif
-}
-#else
-extern int pivot_root(const char * new_root, const char * put_old);
-#endif
 
 #ifdef DEBUG
 #define lxcfs_debug(format, ...)                                               \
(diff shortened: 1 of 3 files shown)


brauner:
I'm unsure how backwards compatible this actually is. Meaning if you have a running instance of `LXCFS` and install a new package including this commit and do a `sudo kill -USR1 $(pidof lxcfs)` this might lead to trouble. A `systemctl restart lxcfs` might be required. Despite that I still like the solution to move the whole cgroup setup code into fuse callbacks.

brauner:
In other words, a `systemctl restart lxcfs` should be done on package upgrade.

stgraber:
@brauner hmm, that'd suck for the stable branch then...

A "systemctl restart lxcfs" breaks all your existing containers, requiring you to restart them all, so not something that'd be suitable for our production users who auto-update to the stable bugfix releases...

brauner:
I fear that if we move anything outside of the `__atttribute(constructor)__` and into `lxcfs.c` then sending `SIGUSR1` to `lxcfs` none of the code we moved into `lxcfs.c` will be re-executed.

brauner:
*Maybe* we can fiddle with `do_reload()`?

stgraber:
@hallyn knows all that stuff much better than I do.

The plan is really that the main C binary be as small as possible, with as little logic as possible. We want bugs and potential security problems to be in the library since that's what we can reload at will without breaking our users.

brauner:
Hm, then I'd argue that we would probably be off best if we leave everything in the constructor even if it isn't the nicest option. Unless @hallyn has a better idea. Moving stuff into `lxcfs.c` isn't so nice anyway since this means we would either have to fuzz around with global variables in a shared library which is not really nice imho  (Especially if you want to keep a consistent view accross all users.) or pass the data around via `fuse`'s `privat_data` but then we still face the problem that anything we mount/setup in `lxcfs.c` can't be reloaded.

hallyn:
Does tests/test_reload.sh pass with this?

brauner:
On Sat, Aug 27, 2016 at 08:51:17PM -0700, Serge Hallyn wrote:
> Does tests/test_reload.sh pass with this?

Sure, it passes. But I think that's also not the problem. The problem ist that
because the setup is now done in lxcfs.c in dedicated fuse lxcfs_init() and
lxcfs_destroy() functions the data (fds, controllers) these two functions
collect and which is needed for the shared library to work after the reload will
*not* be initialized and sent since fuse isn't reloaded. That means when you
simply send SIGUSR1 to lxcfs, then yes, the library will be correctly reloaded
but the cgroup mounts will be busted aka /var/lib/lxcfs/cgroup will not be
correctly setup.

brauner:
Fwiw, we also might get around the problem by not linking `lxcfs` against `liblxcfs`. This way the constructor won't be run before `main`. Rather it would be run, when we first call `dlopen`.

brauner:
Closing in favor of #138.

PR #140, opened 2016-08-30
autotools: add -avoid-version

brauner:


hallyn:
Why?

brauner:
Because we're compiling a libtool module and they are usually unversioned.

Issue #144, opened 2016-09-03
Support running on initramfs


thinxer:
When the system root is on initramfs, `pivot_root` doesn't work and LXCFS fails.

```
pivot_enter: pivot_root() syscall failed: Invalid argument.
```

Docker solves this by providing a `DOCKER_RAMDISK` flag. When set, Docker uses `chroot` instead of `pivot_root`.  Is this possible with lxcfs?

brauner:
That should be doable, yes.

stgraber:
If I remember the kernel code properly, the problem is that you can't pivot_root when your new path is a ramfs. So instead of using env variables, we could probably just statvfs the new root and if it's a ramfs, then use chroot.

Or you could do what I was doing on Android which is to setup a tmpfs, then bind-mount the paths I care about, then pivot_root to that as pivoting to a tmpfs is perfectly fine. The use of the tmpfs can then be made conditional on / being a ramfs.

brauner:
The first option is similar to what we do in `LXC` but for `LXCFS` the second option might make more sense.

brauner:
@hallyn, this can be closed.

PR #145, opened 2016-09-04
lxcfs on ramfs

brauner:
bindings: use chroot() on ramfs  …
- Detect whether we are on a ramfs. We first try via statfs and check for
  RAMFS_MAGIC. This may report TMPFS_MAGIC although it should better report
  RAMFS_MAGIC. In this case, parse /proc/self/mountinfo and check for
        - rootfs rootfs
  like we do in LXC.
- When we are on ramfs use chroot(), otherwise use pivot_root().

According to the kernel code, the current root cannot be on the initramfs when intending to use `pivot_root()`. I went with `chroot()` as I couldn't get it to work with the `tmpfs` option.


brauner:
Closes #144.

PR #152, opened 2016-10-19
fix offsets for memory.stat parsing

Fabian-Gruenbichler:
this seems to have been (wrongly) copy pasted from the old parsing code, which only looked at "total_cache" (which really is 11 chars long ;)).

this especially breaks parsing of the cached memory stat, which means that the "Cached:" value of a container's /proc/meminfo is wrong, which means that tools like  free report cached memory as used instead of cached.


Fabian-Gruenbichler:
just saw that there already is a pull request fixing this, with hard coded lengths - sorry for the noise

PR #155, opened 2016-11-11
pam_cgfs: deal with hybrid cgroupfs mounts

brauner:
When v1 hierarchies are mixed with a v2 hierarchy we currently crash since we
do not handle that case. This adds logic to at least place us in writeable
cgroups and to keep us functional. For now, we simply ignore the v2 hierarchy.
We will add specific logic to deal with it after we know what certain init
systems are planning wrt to v1 vs. v2 cgroup hierarchies.

### DIFF
--- a/pam/pam_cgfs.c
+++ b/pam/pam_cgfs.c
@@ -22,19 +22,20 @@
  * See COPYING file for details.
  */
 
+#include <dirent.h>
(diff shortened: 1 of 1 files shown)


brauner:
I'M closing this in favor of a rewrite, I'm working on.

PR #156, opened 2016-11-14
pam_cgfs: reimplement and add cgroupfs v2 support

brauner:
This is a rewrite of pam_cgfs which leans on LXC's cgfsng.c. Various codepaths
have been adapted and made more appropriate.

The strategy of pam_cgfs v2 is to support cgroupfs v1, cgroupfs v2, and mixed
mounts where some controllers are mounted into a standard cgroupfs v1 hierarchy
location (/sys/fs/cgroup/<controller>) and other controllers are mounted into
the cgroupfs v2 hierarchy.

The functions and types for cgroupfs v1 and cgroupfs v2 have nearly all been
kept separately even if they do nearly the exact same job. This is on purpose!
Although marked non-experimental, cgroupfs v2 is too much of a moving target.
Extrapolating from currentl cgroupfs v2 standard behaviour seems risky and error
prone. Even more so when those assumptions complexify or simplify cgroupfs v1
assumptions when trying to handle both, cgroupfs v1 and cgroupfs v2, in one
function. In short, code duplication currently is on purpose so that we can
easily adapt to changes in cgroupfs v2 behaviour without having to touch any of
the functions or types that deal with the basically standardized cgroupfs v1
behaviour.

A quick run-through of what current pam_cgfs does (The same wording can be found
in the preamble/license to pam_cgfs.c.):

When a user logs in, this pam module will create cgroups which the user may
administer. It handles both pure cgroupfs v1 and pure cgroupfs v2, as well as
mixed mounts, where some controllers are mounted in a standard cgroupfs v1
hierarchy location (/sys/fs/cgroup/<controller>) and others are in the cgroupfs
v2 hierarchy.
Writeable cgroups are either created for all controllers or, if specified, for
any controllers listed on the command line.
The cgroup created will be "user/$user/0" for the first session, "user/$user/1"
for the second, etc.

Systems with a systemd init system are treated specially, both with respect to
cgroupfs v1 and cgroupfs v2. For both, cgroupfs v1 and cgroupfs v2, we check
whether systemd already placed us in a cgroup it created, e.g.

     user.slice/user-uid.slice/session-n.scope

by checking whether uid == our uid. If it did, we simply chown the last
part (session-n.scope). If it did not we create a cgroup as outlined above
(user/$user/n) and chown it to our uid.
The same holds for cgroupfs v2 where checking this assumption becomes crucial:
If we systemd already created and placed us in a cgroups, we __have to__ be
placed our under it on login, otherwise things like starting an xserver or
similar will not work.

All requested cgroups must be mounted under /sys/fs/cgroup/$controller,
no messing around with finding mountpoints.

Note, as of now, we currently do not yet necessarily deal correctly with weird
corner cases like not mounting the name=systemd cgroupfs v1 controller at
/sys/fs/cgroup/systemd but rather mounting an empty cgroupfs v2 hierarchy at the
same location which is used by systemd to track processes. This is left for
future commits.

### DIFF
--- a/pam/pam_cgfs.c
+++ b/pam/pam_cgfs.c
@@ -1161,8 +1161,10 @@ static bool cgv1_init(uid_t uid, gid_t gid)
 		return false;
 
 	f = fopen("/proc/self/mountinfo", "r");
-	if (!f)
+	if (!f) {
+		free(basecginfo);
 		return false;
+	}
 
 	cgv1_get_controllers(&klist, &nlist);
 
@@ -1170,7 +1172,7 @@ static bool cgv1_init(uid_t uid, gid_t gid)
 		char **controller_list = NULL;
 		char *mountpoint, *base_cgroup;
 
-		if (!is_lxcfs(line) && !is_cgv1(line))
+		if (is_lxcfs(line) || !is_cgv1(line))
 			continue;
 
 		controller_list = cgv1_get_proc_mountinfo_controllers(klist, nlist, line);
@@ -1216,7 +1218,15 @@ static bool cgv1_init(uid_t uid, gid_t gid)
 	for (it = cgv1_hierarchies; it && *it; it++) {
 		if ((*it)->controllers) {
 			char *init_cgroup, *user_slice;
+			/* We've already stored the controller and received its
+			 * current cgroup. If we now fail to retrieve its init
+			 * cgroup, we should probably fail.
+			 */
 			init_cgroup = cgv1_get_current_cgroup(basecginfo, (*it)->controllers[0]);
+			if (!init_cgroup) {
+				free(basecginfo);
+				return false;
+			}
 			cg_systemd_prune_init_scope(init_cgroup);
 			(*it)->init_cgroup = init_cgroup;
 			lxcfs_debug("cgroupfs v1 controller \"%s\" has init "
@@ -1400,7 +1410,6 @@ static bool cgv1_enter(const char *cgroup)
 	struct cgv1_hierarchy **it;
 
 	for (it = cgv1_hierarchies; it && *it; it++) {
-		char *path;
 		char **controller;
 		bool entered = false;
 
@@ -1410,6 +1419,7 @@ static bool cgv1_enter(const char *cgroup)
 
 		for (controller = (*it)->controllers; controller && *controller;
 		     controller++) {
+			char *path;
 
 			if ((*it)->systemd_user_slice)
 				continue;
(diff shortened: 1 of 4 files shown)


pam/pam_cgfs.c
```diff
@@ -55,757 +203,1584 @@ static void mysyslog(int err, const char *format, ...)
...
 	size_t len = 0;
-	struct controller *c;
-	bool ret = false;
 
-	f = fopen("/proc/1/cgroup", "r");
+	basecginfo = read_file("/proc/self/cgroup");
+	if (!basecginfo)
+		return false;
+
+	f = fopen("/proc/self/mountinfo", "r");
```
hallyn: Free basecginfo in case of failure?

pam/pam_cgfs.c
```diff
@@ -55,757 +203,1584 @@ static void mysyslog(int err, const char *format, ...)
...
-		}
-		free(subsystems);
-		if (id < 0 || id > 20) {
-			mysyslog(LOG_ERR, "Too many subsystems\n");
-			free(ip);
-			goto out;
+		char **controller_list = NULL;
+		char *mountpoint, *base_cgroup;
+
+		if (!is_lxcfs(line) && !is_cgv1(line))
```
hallyn: Did you really mean *not* is_lxcfs?

pam/pam_cgfs.c
```diff
@@ -55,757 +203,1584 @@ static void mysyslog(int err, const char *format, ...)
...
-		while (c) {
-			fprintf(stderr, " Next mount: index %d name %s\n", c->id, c->name);
-			fprintf(stderr, "             mount path %s\n", c->mount_path ? c->mount_path : "(none)");
-			fprintf(stderr, "             init task path %s\n", c->init_path);
-			fprintf(stderr, "             login task path %s\n", c->cur_path);
-			c = c->next;
+	for (it = cgv1_hierarchies; it && *it; it++) {
+		if ((*it)->controllers) {
+			char *init_cgroup, *user_slice;
+			init_cgroup = cgv1_get_current_cgroup(basecginfo, (*it)->controllers[0]);
```
hallyn: cgv1_get_current_cgroup can return NULL.

brauner:
I would like you to hold off on the merge for a couple of days at least, just in case I detect any errors. :)

stgraber:
Please note that we're looking at Wednesday for LXCFS 2.0.5, if this isn't in by then, then it'll have to wait until 2.0.6 in late January.

hallyn:
Quoting Christian Brauner (notifications@github.com):
> I would like you to hold off on the merge for a couple of days at least, just in case I detect any errors. :)

None of the comments I had were earth-shattering.  So long as you've tested on
a simple v1-only system I think it's good.  Maybe hit it hard and look for
memory and fd leaks :)

Issue #157, opened 2016-11-19
handle cpuset


brauner:
- initialize cgroup.clone_children
- remove isolcpus

brauner:
This is a reminder for myself. @hallyn, please assign me.

brauner:
Closed by #159.

PR #159, opened 2016-11-19
pam_cgfs: handle cgroupfs v1 cpuset controller

brauner:
If init has not placed us into our own cgroup on login we will reside in the
root cgroup. In this case cgroup.clone_children will not have been initialized
and so we need to do it. Otherwise users will not be able to start containers
with cpuset limits set.

### DIFF
--- a/pam/pam_cgfs.c
+++ b/pam/pam_cgfs.c
@@ -518,7 +518,7 @@ static bool mkdir_p(const char *root, char *path)
 			goto next;
 
 		if (mkdir(path, 0755) < 0) {
-			lxcfs_debug("Failed to create %s: %m.\n", path);
+			lxcfs_debug("Failed to create %s: %s.\n", path, strerror(errno));
 			return false;
 		}
 
@@ -575,13 +575,13 @@ next:
 
 	if (rmdir(dirname) < 0) {
 		if (!r)
-			lxcfs_debug("Failed to delete %s: %m.\n", dirname);
+			lxcfs_debug("Failed to delete %s: %s.\n", dirname, strerror(errno));
 		r = -1;
 	}
 
 	if (closedir(dir) < 0) {
 		if (!r)
-			lxcfs_debug("Failed to delete %s: %m.\n", dirname);
+			lxcfs_debug("Failed to delete %s: %s.\n", dirname, strerror(errno));
 		r = -1;
 	}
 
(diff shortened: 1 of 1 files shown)


brauner:
Closes #157.

brauner:
Fwiw, we should try to land this commit soon before the release on Wednesday because its crucial for the cpuset hierarchy to be usable for unprivileged containers. :) Also, if a few more people are willing to test and hit this hard this would be great. ;)

hallyn:
Sorry, I'm not understanding.  Can you please give a detailed example situation?

brauner:
Sure. Assume a `systemd`-based distro which does *not* use `pam_cgfs`. `systemd` might decide to put us into the root cgroup for the `cpuset` controller. Which would look like this:

```
chb@conventiont|~
> cat /proc/self/cgroup
11:cpuset:/
10:blkio:/user.slice
9:perf_event:/
8:memory:/user.slice
7:net_cls,net_prio:/
6:devices:/user.slice
5:hugetlb:/
4:cpu,cpuacct:/user.slice
3:pids:/user.slice/user-1000.slice/session-1.scope
2:freezer:/
1:name=systemd:/user.slice/user-1000.slice/session-1.scope
```

In this case `/sys/fs/cgroup/cpuset/cgroup.clone_children` will be set to `0`:

```
chb@conventiont|~
> cat /sys/fs/cgroup/cpuset/cgroup.clone_children
0
```

And so when we `mkdir_p("/sys/fs/cgroup/cpuset/user/$user/0/")` we wouldn't be able to join the `cpuset` controller even though we have created a writeable cgroup for us, since `cgroup.clone_children` is `0` in the root cgroup and hence, we cannot give us any cpus in our writeable cgroup.

pam/pam_cgfs.c
```diff
@@ -1558,6 +1602,438 @@ static bool cg_belongs_to_uid_gid(const char *path, uid_t uid, gid_t gid)
...
+	if (read_from_file(clonechildrenpath, &v, 1) < 0) {
+		lxcfs_debug("Failed to read '%s'", clonechildrenpath);
+		free(clonechildrenpath);
+		free(cgpath);
+		return false;
+	}
+
+	/* Make sure any isolated cpus are removed from cpuset.cpus. */
+	if (!cg_filter_and_set_cpus(cgpath, v == '1'))
+		return false;
```
hallyn: No freeing here?

PR #162, opened 2016-12-22
fix swap values with nested cgroups

Fabian-Gruenbichler:
the memory limit was already correctly set by looking at the
whole cgroup hierarchy and using the minimum value, refactor
that code to support arbitrary files in the memory cgroup
and reuse it for the memsw limit as well.

Signed-off-by: Fabian Grünbichler <f.gruenbichler@proxmox.com>

### DIFF
--- a/bindings.c
+++ b/bindings.c
@@ -3046,12 +3046,12 @@ static int read_file(const char *path, char *buf, size_t size,
  * FUSE ops for /proc
  */
 
-static unsigned long get_memlimit(const char *cgroup)
+static unsigned long get_memlimit(const char *cgroup, const char *file)
 {
 	char *memlimit_str = NULL;
 	unsigned long memlimit = -1;
 
-	if (cgfs_get_value("memory", cgroup, "memory.limit_in_bytes", &memlimit_str))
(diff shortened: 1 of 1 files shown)


Fabian-Gruenbichler:
there are two remaining questions for me:
- the check whether swap accounting is enabled still uses the old style (non minimum-recursive) way to get the memsw limit, so this might break in some scenarios (but in those it would be broken now as well)?
- the default limit and usage chunks which I removed did not make sense to me at all, and the commits introducing them don't really give a lot of background. I did some tests using both regular one cgroup style setup and the nested cgroup as introduced with https://github.com/lxc/lxc/pull/1302 , and it seems to work without them..

I did not test with lxd and nested containers so far, but don't expect problems there.

Fabian-Gruenbichler:
and because I just realized that this is not obvious from the changes and the commit message, without this commit lxcfs will use the host's swap limit and usage for /proc/meminfo and /proc/swaps when the container is setup with the "inner" cgroup as described in https://github.com/lxc/lxc/pull/1302

hallyn:
Looks good to me, thanks.

PR #168, opened 2017-01-27
Also check next variable for NULL on cg_rmdir

Xeno22:
lxcfs seems to crash in cg_rmdir while phpsessioncleanup is running in many containers simultaneously.

Here is a stack trace:
```
Detaching after fork from child process 17921.
Reading in symbols for ../sysdeps/x86_64/multiarch/../strcmp.S...done.

Program received signal SIGSEGV, Segmentation fault.
[Switching to Thread 0x7fba74fc4700 (LWP 214112)]
__strcmp_ssse3 () at ../sysdeps/x86_64/multiarch/../strcmp.S:1358
1358    ../sysdeps/x86_64/multiarch/../strcmp.S: No such file or directory.
Reading in symbols for ../sysdeps/unix/sysv/linux/x86_64/clone.S...done.
#0  __strcmp_ssse3 () at ../sysdeps/x86_64/multiarch/../strcmp.S:1358
No locals.
#1  0x00007fba75099de3 in cg_rmdir () from /usr/lib/lxcfs/liblxcfs.so
No symbol table info available.
#2  0x000055952dda6c8e in ?? ()
No symbol table info available.
#3  0x00007fba75a9a404 in ?? () from /lib/x86_64-linux-gnu/libfuse.so.2
No symbol table info available.
#4  0x00007fba75aa422b in ?? () from /lib/x86_64-linux-gnu/libfuse.so.2
No symbol table info available.
#5  0x00007fba75aa0e49 in ?? () from /lib/x86_64-linux-gnu/libfuse.so.2
No symbol table info available.
#6  0x00007fba75675184 in start_thread (arg=0x7fba74fc4700) at pthread_create.c:312
```

I can only guess that the caller_is_in_ancestor returns false and sets next to NULL - then the strcmp crashes.

I hope the ret errno is correct for this case.

### DIFF
--- a/bindings.c
+++ b/bindings.c
@@ -2908,7 +2908,7 @@ int cg_rmdir(const char *path)
 	if (initpid <= 0)
 		initpid = fc->pid;
 	if (!caller_is_in_ancestor(initpid, controller, cgroup, &next)) {
-		if (!last || strcmp(next, last) == 0)
+		if (!last || (next && (strcmp(next, last) == 0)))
 			ret = -EBUSY;
 		else
 			ret = -ENOENT;


stgraber:
Looks like you'll want to rebase your branch and squash your commits together to have a single commit including the Signed-off-by line.

stgraber:
jenkins: test this please

stgraber:
@brauner @hallyn

brauner:
@stgraber can't merge. Afaik, I don't yet have write access.

stgraber:
@brauner yeah, was mostly wondering if that made sense to you or whether this is papering over a bigger problem that should be fixed at the source.

Xeno22:
It could be a bigger issue, but at least it checks for the invalid use of NULL in strcmp.
Maybe there is some race condition.

brauner:
AFAICT, no since  this function may validly return null and advertises this fact in its comment. I may however spend some time rewriting parts of it after the storage stuff is sorted in preparation for the big cgroup v2 translation problem...

brauner:
Parts of bindings.c that is

hallyn:
Note that '!last' cannot happen there, as we exit earlier in the fn if last is NULL.  So I wonder whether that !last was meant to be !next all along.

Xeno22:
Should the fn return ENOENT or EBUSY if `!caller_is_in_ancestor() && next = NULL`?

hallyn:
I think it should be ENOENT.

brauner:
If you make that change, please test, that standard operations like `rmdir`, `mkdir`, `rm`, and `touch` behave as expected under the cgroup part of the `lxcfs` mountpoint. This can be quite "delicate".

brauner:
@Xeno22 that is. :)

PR #169, opened 2017-01-30
virtualize the 'btime' field of /proc/stat

almostivan:
Currently, the 'btime' of /proc/stat reflects the boot time of the host.
We would like it to reflect when the guest boots, so use the start time of
init.

Signed-off-by: Jason Baron <jbaron@akamai.com>


hallyn:
Thanks!

PR #174, opened 2017-02-09
2017 02 09/bugfixes

brauner:


pam/pam_cgfs.c
```diff
@@ -1650,7 +1650,7 @@ static char *string_join(const char *sep, const char **parts, bool use_as_prefix
 	size_t result_len = use_as_prefix * sep_len;
 
 	/* calculate new string length */
-	for (p = (char **)parts; *p; p++)
+	for (p = (char **)parts; p && *p; p++)
```
hallyn: Thanks, everything looks good except for this one.  This can only happen if parts is NULL coming in, right?  Can you just return at the top of the fn if parts is NULL?

brauner: yup

PR #182, opened 2017-05-16
pam: bugfixes

brauner:
When we detected the unified cgroup hierarchy we actually need to report back that we did. Otherwise we might needlessly fail later on.

### DIFF
--- a/pam/pam_cgfs.c
+++ b/pam/pam_cgfs.c
@@ -1389,6 +1389,7 @@ static bool cgv2_init(uid_t uid, gid_t gid)
 		 * each of those mountpoints will expose identical information.
 		 * So let the first mountpoint we find, win.
 		 */
+		ret = true;
 		break;
 	}
 
--- a/pam/pam_cgfs.c
+++ b/pam/pam_cgfs.c
@@ -2250,7 +2250,7 @@ static bool cgv2_create(const char *cgroup, uid_t uid, gid_t gid, bool *existed)
 	char *clean_base_cgroup;
 	char *path;
 	struct cgv2_hierarchy *v2;
-	bool created = false;
+	bool our_cg = false, created = false;
 
 	*existed = false;
 
@@ -2262,10 +2262,9 @@ static bool cgv2_create(const char *cgroup, uid_t uid, gid_t gid, bool *existed)
 	/* We can't be placed under init's cgroup for the v2 hierarchy. We need
 	 * to be placed under our current cgroup.
 	 */
-	if (cg_systemd_chown_existing_cgroup(v2->mountpoint,
-				v2->base_cgroup, uid, gid,
-				v2->systemd_user_slice))
-		return true;
+	if (cg_systemd_chown_existing_cgroup(v2->mountpoint, v2->base_cgroup,
+					     uid, gid, v2->systemd_user_slice))
+		goto chown_cgroup_procs_file;
 
 	/* We need to make sure that we do not create an endless chain of
 	 * sub-cgroups. So we check if we have already logged in somehow (sudo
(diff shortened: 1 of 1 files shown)


brauner:
This doesn't deal with enabling controllers for user just now. I'd like to think through this first.
I also suspect that what we want to do soon is to create an additional sub-cgroup for the user. Meaning:

```
    /sys/fs/cgroup/c1/c2
  /
c3
```
where

    /sys/fs/cgroup/c1/c2
    /sys/fs/cgroup/c1/c2/cgroup.procs

is owned by our unprivileged user and we've created

    /sys/fs/cgroup/c1/c2/c3

by doing something along the lines of

    getresuid(u1, u2, u3);
    getresgid(g1, g2, g3);
    setresuid(unpriv-uid, unpriv-uid, -1);
    setresgid(unpriv-gid, unpriv-gid, -1);
    mkdir("/sys/fs/cgroup/c1/c2/c3", 0755);
    setresuid(u1, u2, u3);
    setresgid(g1, g2, g3);

brauner:
```
chb@conventiont|~
> cat /proc/self/cgroup
11:pids:/user.slice/user-1000.slice/session-4.scope
10:cpuset:/
9:memory:/user/chb/0
8:perf_event:/
7:blkio:/user.slice
6:net_cls,net_prio:/
5:devices:/user.slice
4:cpu,cpuacct:/user.slice
3:hugetlb:/
2:freezer:/user/chb/0
1:name=systemd:/user.slice/user-1000.slice/session-4.scope
0::/user.slice/user-1000.slice/session-4.scope
```
and if `unified` is specified in the `common-session` line for `pam_cgfs.so`:

```
chb@conventiont|~
> ls -al /sys/fs/cgroup/unified/user.slice/user-1000.slice/session-4.scope/
total 0
drwxr-xr-x 2 chb  chb  0 May 16 17:36 .
drwxr-xr-x 5 root root 0 May 16 17:36 ..
-r--r--r-- 1 root root 0 May 16 17:36 cgroup.controllers
-r--r--r-- 1 root root 0 May 16 17:36 cgroup.events
-rw-r--r-- 1 chb  chb  0 May 16 17:36 cgroup.procs
-rw-r--r-- 1 root root 0 May 16 17:36 cgroup.subtree_control
```

pam/pam_cgfs.c
```diff
@@ -2291,7 +2290,7 @@ static bool cgv2_create(const char *cgroup, uid_t uid, gid_t gid, bool *existed)
 			*existed = false;
 		else
 			*existed = true;
-		return our_cg;
+		goto chown_cgroup_procs_file;
```
hallyn: our_cg could have been false, but you're jumping to a block which unconditionally returns true.

pam/pam_cgfs.c
```diff
@@ -2300,10 +2299,27 @@ static bool cgv2_create(const char *cgroup, uid_t uid, gid_t gid, bool *existed)
...
+	/* chown cgroup to user */
 	if (chown(path, uid, gid) < 0)
 		mysyslog(LOG_WARNING, "Failed to chown %s to %d:%d: %s.\n",
 			 path, (int)uid, (int)gid, strerror(errno), NULL);
-	lxcfs_debug("Chowned %s to %d:%d.\n", path, (int)uid, (int)gid);
+	else
+		lxcfs_debug("Chowned %s to %d:%d.\n", path, (int)uid, (int)gid);
+	free(path);
+
+chown_cgroup_procs_file:
```
hallyn: The fact that 'path' is reassigned here becomes worrisome - i wonder whether this block should be a different function.  Not asking for that for this patch.

pam/pam_cgfs.c
brauner: Thanks!

pam/pam_cgfs.c
brauner: Yeah, I suspect I'll refactor this once I'll do the sub-cgroup thing.

pam/pam_cgfs.c
brauner: fixed

brauner:
@hallyn, updated.

hallyn:
Sorry, it still looks like you are returning true if our_cg is false, whereas before you would have returned false in that case.  I might be misunderstanding the flow at this point, but if the cg prevoiusly existed and does not belong to us, don't you just want to return false with existed=true at that point?  In your case if !our_cg you proceed to try to mkdir_p(v2->mountpoint, path) ?

brauner:
Right. Updated.

PR #183, opened 2017-05-17
bindings: add mountpoint for unified hierarchy

brauner:

### DIFF
--- a/bindings.c
+++ b/bindings.c
@@ -4383,11 +4383,13 @@ static bool cgfs_mount_hierarchies(void)
 
 	for (i = 0; i < num_hierarchies; i++) {
 		char *controller = hierarchies[i];
+
 		clen = strlen(controller);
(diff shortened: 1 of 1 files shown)


bindings.c
```diff
@@ -4449,18 +4455,20 @@ static void __attribute__((constructor)) collect_and_mount_subsystems(void)
...
 	}
 
 	while (getline(&line, &len, f) != -1) {
-		char *p, *p2;
+		char *idx, *p, *p2;
 
 		p = strchr(line, ':');
 		if (!p)
 			goto out;
+		idx = line;
```
hallyn: This might be clearer so might be worth it anyway, but just to check - is there any reason you couldn't just use 'line' instead of idx in the strcmp below?

brauner: Nah, it was really just about naming it `idx` but I can remove it. :)

PR #186, opened 2017-05-26
bindings: calculate uptime via proc/<pid>/stat

### LINKED ISSUE #165: lxcfs 2.0.5 /proc/uptime always resets to 0
Hello there,

I've just noticed that the containers uptime is always reported as "0 min", so I did a further investigation and discovered that /proc/uptime behaves really strange.

Host distro is Debian 7 (to not mess with systemd's intelligent I-do-everything-for-you behavior), LXD version is 2.7 compiled from github sources (as of 13.01.2017) using go-1.6.4, lxc-lib is 2.0.6, kernel is taken from Yakkety 4.8.0-34.36 (unmodified), and we're using the stock lxcfs 2.0.5 mount hook script for the containers (unmodified).

There's no error / warning upon containers start, everything looks fine. Cgroups memory virtualization looks fine inside containers. The only warning that we get upon container start is from dmesg's output saying "cgroup: new mount options do not match the existing superblock, will be ignored", but I assume that's just safe to ignore (maybe I'm wrong).

The problem is as follows:

`root@finaltest:~# lxc start ubuntu`
`root@finaltest:~# lxc exec ubuntu bash`
`root@ubuntu:~# w`
` 17:07:36 up 0 min,  0 users,  load average: 0.18, 0.09, 0.04`
`root@ubuntu:~# while true; do sleep 5s; cat /proc/uptime; w | grep up; done`
`41.0 41.0`
` 17:10:55 up 0 min,  0 users,  load average: 0.10, 0.08, 0.04`
`46.0 46.0`
` 17:11:00 up 0 min,  0 users,  load average: 0.09, 0.08, 0.04`
`0.0 0.0`
` 17:11:05 up 0 min,  0 users,  load average: 0.08, 0.08, 0.04`
`5.0 5.0`
` 17:11:10 up 0 min,  0 users,  load average: 0.08, 0.08, 0.04`
`10.0 10.0`
` 17:11:15 up 0 min,  0 users,  load average: 0.07, 0.08, 0.04`
`15.0 15.0`
` 17:11:20 up 0 min,  0 users,  load average: 0.07, 0.08, 0.04`
`20.0 20.0`
` 17:11:25 up 0 min,  0 users,  load average: 0.06, 0.07, 0.04`
`25.0 25.0`
` 17:11:30 up 0 min,  0 users,  load average: 0.06, 0.07, 0.04`
`30.0 30.0`
` 17:11:35 up 0 min,  0 users,  load average: 0.05, 0.07, 0.04`
`35.0 35.0`
` 17:11:40 up 0 min,  0 users,  load average: 0.05, 0.07, 0.04`
`40.0 40.0`
` 17:11:45 up 0 min,  0 users,  load average: 0.04, 0.07, 0.04`
`45.0 45.0`
` 17:11:50 up 0 min,  0 users,  load average: 0.04, 0.07, 0.04`
`50.0 50.0`
` 17:11:55 up 0 min,  0 users,  load average: 0.04, 0.07, 0.04`
`55.0 55.0`
` 17:12:00 up 0 min,  0 users,  load average: 0.03, 0.06, 0.03`
`0.0 0.0`
` 17:12:05 up 0 min,  0 users,  load average: 0.03, 0.06, 0.03`
`^C`

You can clearly see the first value never gets above 50-55 seconds, sometimes it even resets to zero after reaching 40-45 seconds.

So I've looked at lxcfs source and discovered that the `getreaperage()` function is responsible for virtualizing container's /proc/uptime. Seems like `lookup_initpid_in_store()` is somehow not able to correctly determine the container's PID 1 process directory outside the container (my assumption) and hence this problem. Of course it could be something else.

I would really like to have the latest LXD + LXC + LXCFS running smoothly on Wheezy, please let me know if I can give you some more information to investigate this problem further. Together we can make this work!

Thank you!

### LINKED ISSUE #184: Proposal: Calculate uptime for container with start time of init process in container
Now, we use  [/proc/$init_pid](https://github.com/lxc/lxcfs/blob/master/bindings.c#L3475) 's ctime as container's uptime.  However, when kernel drop inode cache, such as run `echo 3 > /proc/sys/vm/drop_caches`, will cause `/proc/$init_pid` inode newly created, this will make uptime in container become zero. @hallyn

brauner:
Closes #165.
Closes #184.

### DIFF
--- a/bindings.c
+++ b/bindings.c
@@ -8,14 +8,17 @@
 
 #define FUSE_USE_VERSION 26
 
+#define __STDC_FORMAT_MACROS
 #include <dirent.h>
 #include <errno.h>
 #include <fcntl.h>
 #include <fuse.h>
+#include <inttypes.h>
 #include <libgen.h>
 #include <pthread.h>
 #include <sched.h>
 #include <stdbool.h>
+#include <stdint.h>
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
@@ -30,10 +33,14 @@
 #include <sys/param.h>
 #include <sys/socket.h>
 #include <sys/syscall.h>
+#include <sys/sysinfo.h>
 #include <sys/vfs.h>
 
 #include "bindings.h"
 #include "config.h" // for VERSION
+
+/* Maximum number for 64 bit integer is a string with 21 digits: 2^64 - 1 = 21 */
+#define LXCFS_NUMSTRLEN64 21
 
 /* Define pivot_root() if missing from the C library */
 #ifndef HAVE_PIVOT_ROOT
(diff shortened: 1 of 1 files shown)


bindings.c
```diff
@@ -3454,25 +3461,167 @@ static int proc_cpuinfo_read(char *buf, size_t size, off_t offset,
...
 
 	qpid = lookup_initpid_in_store(pid);
-	if (qpid <= 0)
+	if (qpid <= 0) {
+		/* Caller can check for EINVAL on 0. */
+		errno = EINVAL;
+		return 0;
+	}
+
+	ret = snprintf(path, MAXPATHLEN, "/proc/%d/stat", qpid);
```
hallyn: path isn't that big though.

bindings.c
```diff
@@ -3454,25 +3461,167 @@ static int proc_cpuinfo_read(char *buf, size_t size, off_t offset,
...
+			"%*d "      /* (17) cstime      %ld  */
+			"%*d "      /* (18) priority    %ld  */
+			"%*d "      /* (19) nice        %ld  */
+			"%*d "      /* (20) num_threads %ld  */
+			"%*d "      /* (21) itrealvalue %ld  */
+			"%" PRIu64, /* (22) starttime   %llu */
+		     &starttime);
+	if (ret != 1) {
+		/* Caller can check for EINVAL on 0. */
+		errno = EINVAL;
```
hallyn: leaking open f

brauner: Ah, because I changed the fun again right before pushing, damn.

bindings.c
```diff
@@ -3454,25 +3461,167 @@ static int proc_cpuinfo_read(char *buf, size_t size, off_t offset,
...
+
+	ticks_per_sec = sysconf(_SC_CLK_TCK);
+	if (ticks_per_sec < 0 && errno == EINVAL) {
+		lxcfs_debug("%s\n", "failed to determine number of clock ticks in a second");
+		return 0;
+	}
 
-	if (lstat(fnam, &sb) < 0)
+	uptime = clockticks;
+	uptime /= ticks_per_sec;
```
hallyn: Heh, I wonder if someone could mess with the env so as to make ticks_per_sec = 0.

brauner: `sysconf()` well, not sure. But they would have to do it at compile time, no?

hallyn:
On Thu, May 25, 2017 at 09:43:55PM -0700, Christian Brauner wrote:
> brauner commented on this pull request.
> 
> 
> 
> >  
> -	if (lstat(fnam, &sb) < 0)
> +	uptime = clockticks;
> +	uptime /= ticks_per_sec;
> 
> `sysconf()` well, not sure. But they would have to do it at compile time, no?

LD_PRELOAD maybe? Not sure.  Just a thought.

brauner:
I think we'll just rely on `CLOCKS_PER_SEC` thereby avoiding sysconf.

hallyn:
Awesome, thanks!

PR #188, opened 2017-06-08
Add reload in service scripts

theredcat:
Add reload feature in sysV and systemd scripts.

Upstart doesn't have such an option since it assume that every service use the SIGHUP for configuration reload

### DIFF
--- a/config/init/sysvinit/lxcfs
+++ b/config/init/sysvinit/lxcfs
@@ -80,6 +80,13 @@ case "$1" in
         fi
         status_of_proc -p ${PIDFILE} "${DAEMON}" lxcfs
     ;;
+    
+    reload)
+        if init_is_upstart; then
+            exit 1
+        fi
+        kill -USR1 $(cat /var/run/lxcfs.pid)
+    ;;
 
     restart|force-reload)
         if init_is_upstart; then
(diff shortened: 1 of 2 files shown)


config/init/sysvinit/lxcfs
```diff
@@ -80,6 +80,13 @@ case "$1" in
         fi
         status_of_proc -p ${PIDFILE} "${DAEMON}" lxcfs
     ;;
+    
+    reload)
+        if init_is_upstart; then
+            exit 1
+        fi
+        kill -USR1 $(cat /var/run/lxcfs.pid)
```
evgeni: This should be `kill -USR1 $(cat ${PIDFILE})`

evgeni:
* I think this should go to master, and then be backported?
* Is it really safe to send USR1? What happens to running containers?

stgraber:
Correct, this branch is lacking the necessary Signed-off-by line in the commit message (per CONTRIBUTING) and needs to be targeted to the master branch, we'll then cherry-pick this into stable-2.0.

As for the signal handling, lxcfs is split into a shim binary and a library. That's so we can update it without having to respawn lxcfs itself, which would otherwise cause /proc in all existing containers to stop working. SIGUSR1 is what tells lxcfs that a new version of the library is available and that it should open it and use it as soon as it's done with current requests.

stgraber:
@theredcat 

Hey there,

Can you re-submit this against master, resolve @evgeni's comment and make sure your commits include the "Signed-off-by:" line?

With that done, we can then merge your change.

Thanks!

brauner:
Closed by https://github.com/lxc/lxcfs/pull/188 .

PR #196, opened 2017-06-28
enable debug option

elianka:
add a switch "-d" to lxcfs. its useful when debugging libfuse.


brauner:
Thanks!

PR #197, opened 2017-06-30
modify default liblxcfs.so path, becasue the default prefix is /usr/l…

elianka:
…ocal, and liblxcfs.so will be installed in /usr/local/lib/ by default.

image

### DIFF
--- a/lxcfs.c
+++ b/lxcfs.c
@@ -86,7 +86,7 @@ static void do_reload(void)
 		goto good;
 	}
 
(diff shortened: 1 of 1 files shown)


hallyn:
Hm.  Note that first we try to dlopen without a path, so ld.so should find it regardless.  Really this fallback doesn't make much sense.  If we're going to keep it then we should really have autoconf provide the fallback pathname.  Can you change this patch to do that?

elianka:
all right. i will do it.
close it for more common fallback.

PR #198, opened 2017-07-01
add common fallback dlopen for liblxcfs.so

elianka:
There is definition for libdir in configure.ac.
`AS_AC_EXPAND(LIBDIR, "$libdir")`
so we could use LIBDIR in our code.
image


hallyn:
Thanks!

PR #199, opened 2017-07-12
modify the error lib path

elianka:

### DIFF
--- a/lxcfs.c
+++ b/lxcfs.c
@@ -90,9 +90,9 @@ static void do_reload(void)
 
 #ifdef LIBDIR
 	/* LIBDIR: autoconf will setup this MACRO. Default value is $PREFIX/lib */
-        snprintf(lxcfs_lib_path, PATH_MAX, "%s/lxcfs/liblxcfs.so", LIBDIR);
+        snprintf(lxcfs_lib_path, PATH_MAX, "%s/liblxcfs.so", LIBDIR);
 #else
-        snprintf(lxcfs_lib_path, PATH_MAX, "/usr/local/lib/lxcfs/liblxcfs.so");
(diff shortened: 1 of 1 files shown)


elianka:
modify the error commit in #198

hallyn:
lxcfs.so in fact ends up in LIBDIR/lxcfs/ does it not?

stgraber:
Yeah, I'm not sure what the problem is here. It's very intentional that liblxcfs isn't in /lib directly but in a subdirectory as it's an internal component of lxcfs which shouldn't be linked against by anyone else.

brauner:
Note also that a while ago I made `liblxcfs` a versionless `libtool` module. So linking against it becomes even more pointless.

elianka:
@hallyn liblxcfs.so is located at LIBDIR not LIBDIR/lxcfs/

elianka:
@stgraber actually, now the liblxcfs.so is not in a subdirectory in /lib. the default path is /usr/local/lib/liblxcfs.so now.

elianka:
@brauner 
this is the tradeoff library path for liblxcfs.so, if we don't set LD_LIBRARY_PATH, we always can not find it. so the fallback dlopen path is usable. 
i have modified it in #198 , meanwhile, i found that the default path is LIBDIR, not LIBDIR/lxcfs. ++

brauner:
But that is caused by compile instructions is it not? The things it that `liblxcfs.so` should stay under a private directory `lxcfs`. This is best practice for `libtool` modules and prevents other programs from trying to link against it.

elianka:
@brauner i fully agree with you about the private directory, meanwhile, it is not now, the default lib path is /usr/local/lib/liblxcfs.so. so maybe a patch should be added.

stgraber:
Right now, our code DOES look at "lxcfs/liblxcfs.so", it's your branch which reverts this very behavior.

Closing this.

PR #220, opened 2017-10-20
Fix wrong calc of swaptoal and swapfree

yuwang888:
it make the value of (memswlimit - memlimit) to be swaptotal,
it is wrong, because swapsize in cgroup/container can up to
[0,memswlimit], sometimes if the memsize(unless swap) of all tasks in
cgroup/container is very small, the swaptoal can to be memswlimit
so make the swaptotal to be min(host swtoal,memswlimit)

Signed-off-by: yuwang.yuwang <yuwang.yuwang@alibaba-inc.com>

### DIFF
--- a/bindings.c
+++ b/bindings.c
@@ -3169,12 +3169,12 @@ static int proc_meminfo_read(char *buf, size_t size, off_t offset,
 			printme = lbuf;
 		} else if (startswith(line, "SwapTotal:") && memswlimit > 0) {
 			sscanf(line+sizeof("SwapTotal:")-1, "%lu", &hostswtotal);
-			if (hostswtotal < memswlimit - memlimit)
-				memswlimit = hostswtotal + memlimit;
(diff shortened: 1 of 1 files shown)


brauner:
This seems correct to me. @Blub, you were the one to add this line before. Do you think this is reasonable?

Blub:
Looks like I was stuck in the mem vs swap vs memsw hell during that last fixup.
Ack.

brauner:
We've all been there...

time-river:
It makes me confused.
There is still [`swap_total = (memswlimit - memlimit) / 1024;`](https://github.com/lxc/lxcfs/blob/b3aa7f1a115626f263f019ff6d8ef54f0132f025/bindings.c#L5366) in v3.1.2.
According to [Linux doc](https://www.kernel.org/doc/Documentation/cgroup-v1/memory.txt):

>  memory.limit_in_bytes		 # set/show limit of memory usage
> memory.memsw.limit_in_bytes	 # set/show limit of memory+Swap usage

So, which is right? @Blub @brauner

PR #221, opened 2017-11-02
pam: cgroup v2 fixes

brauner:
In order to enable proper unprivileged cgroup delegation on newer kernels we not
just need delegate the "cgroup.procs" file to the user but also to
"cgroup.threads". But don't report an error in case it doesn't exist.
Also delegate "cgroup.subtree_control" to enable users to hand over controllers
to descendant cgroups.

### DIFF
--- a/bindings.c
+++ b/bindings.c
@@ -3593,24 +3593,6 @@ static uint64_t get_reaper_age(pid_t pid)
 	return procage;
 }
 
-static uint64_t get_reaper_btime(pid)
-{
-	int ret;
-	struct sysinfo sys;
-	uint64_t procstart;
-	uint64_t uptime;
-
-	ret = sysinfo(&sys);
-	if (ret < 0) {
-		lxcfs_debug("%s\n", "failed to retrieve system information");
-		return 0;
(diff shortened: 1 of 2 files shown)


bindings.c
```diff
@@ -4477,6 +4479,19 @@ static bool permute_root(void)
 	return true;
 }
 
+static int preserve_ns(int pid)
```
hallyn: This could be confusing later, could you name it preserve_mnt_ns() for precision?

bindings.c
```diff
@@ -4477,6 +4479,19 @@ static bool permute_root(void)
...
+{
+	int ret;
+	size_t len = 5 /* /proc */ + 21 /* /int_as_str */ + 7 /* /ns/mnt */ + 1 /* \0 */;
```
hallyn: Any reason not to do "sizeof("/proc") + sizeof("/ns/mnt") + ...?

brauner: No, can do.

brauner:
@hallyn, should be good to got if you're fine with the changes.

PR #224, opened 2017-11-28
Removing the empty README file

abbasally5:
- There is already a README.md that is being used

Signed-off-by: Abbas Ally <abbasally5@yahoo.com>


brauner:
The `README` file is required by autotools. Otherwise they won't compile. :)

abbasally5:
can I add a line to the Makefile such as
README: README.md
and add it to the files to be cleaned?
Sorry for the small change. I a trying to get into open source

brauner:
I'd prefer to just leave it actually. We have this in other autotools based projects as well. It's not a big deal.

abbasally5:
Okay no problem.

PR #225, opened 2017-12-02
Fix inaccurate values in /proc/meminfo for containers with child cgroups

asokoloski:
The values for Cached, Active, Inactive, Active(anon), Inactive(anon),
Active(file), Inactive(file), and Unevictable are derived/computed
from these values in the relevant meminfo.stat:

cache
active_anon
inactive_anon
active_file
inactive_file
unevictable

However, these value apply only to the cgroup of the lxc container
itself.  If your container uses memory cgroups internally, and thus
the container cgroup has children, their memory is not counted.

In order to take the memory usage of child cgroups into account, we
need to look at the "total_" prefixed versions of these values.

Signed-off-by: Aaron Sokoloski <asokoloski@gmail.com>

### DIFF
--- a/bindings.c
+++ b/bindings.c
@@ -2961,23 +2961,23 @@ static void parse_memstat(char *memstat, unsigned long *cached,
 	char *eol;
 
 	while (*memstat) {
-		if (startswith(memstat, "cache")) {
-			sscanf(memstat + 5, "%lu", cached);
+		if (startswith(memstat, "total_cache")) {
+			sscanf(memstat + 11, "%lu", cached);
 			*cached /= 1024;
-		} else if (startswith(memstat, "active_anon")) {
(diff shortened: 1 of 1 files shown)


asokoloski:
I've looked through the open issues, but I don't believe that any of them relate to this problem.

Thanks for making some very useful software!

brauner:
Nicely spotted. Thanks!

tebanep:
Bumped into this issue as well when realizing that some containers where swapping without needing it since the host reported enough free memory. Do you think a 2.0.9 release, including this fix, will happen any time soon? Thank you!

hallyn:
@tebanep I just checked, it is in stable-2.0 tag.

tebanep:
Yes @hallyn I noticed the same, thank you! I would like to know, though, if a new minor release is close.

PR #226, opened 2017-12-04
Change MemAvailable figure in /proc/meminfo to include cache memory -- Fixes #175 I think.

### LINKED ISSUE #175: Incorrect RAM usage report in containers
Hello,

I have a problem with lxcfs ram report in containers. It seems that the cached files from disk is shown as used memory.

Here is a example with a little one liner that :

- List all containers on the system
- Lxc-attach and exec `free -m` and grep `used - buffers/cache` (aka : the real used memory)
- Sum

Then a `free -m` with the same value but on the host.

```
14:12:56 root@hypervisor-01:~ # for i in $(lxc-ls -1); do lxc-attach -n $i -- free -m|grep 'buffers/cache' |awk '{print $3}'; done| awk '{s+=$1} END {printf "%.0f\n", s}'
97365
14:12:59 root@hypervisor-01:~ # free -m |grep 'buffers/cache' |awk '{print $3}'
46074
14:14:06 root@hypervisor-01:~ # lxc-ls --version
2.0.5
14:14:09 root@hypervisor-01:~ # lxcfs --version                                                                                                                                            
2.0.5
14:14:13 root@hypervisor-01:~ #
```

asokoloski:
MemAvailable represents roughly how much more memory we can use before
we start swapping.  Page cache memory can be reclaimed if it's needed
for something else, so it should count as available memory.  This
change should also fix the "available" column of the "free" command,
as well as the "avail Mem" value in "top", both of which come from
MemAvailable.

Note that this isn't perfectly accurate.  On a physical machine, the
value for MemAvailable is the result of a calculation that takes into
account that when memory gets low (but before it's completely
exhausted), kswapd wakes up and starts paging things out.  See:

https://github.com/torvalds/linux/blob/a0908a1b7d68706ee52ed4a039756e70c8e956e9/mm/page_alloc.c#L4553
(si_mem_available function)

I tried to think of a way to be more exact, but this calculation
includes figures that we don't have available for a given cgroup
hierarchy, such as reclaimable slab memory and the low watermark for
zones.  So it's not really feasible to reproduce it exactly.

But anyway, since the kernel calculation itself is just an estimation,
it doesn't seem too bad that we're a little bit off.  Adding in the
amount of memory used for page cache seems much better than what we
were doing before (just copying the free memory figure), because that
can be wrong by gigabytes.

### DIFF
--- a/bindings.c
+++ b/bindings.c
@@ -3167,7 +3167,7 @@ static int proc_meminfo_read(char *buf, size_t size, off_t offset,
 			snprintf(lbuf, 100, "MemFree:        %8lu kB\n", memlimit - memusage);
 			printme = lbuf;
 		} else if (startswith(line, "MemAvailable:")) {
-			snprintf(lbuf, 100, "MemAvailable:   %8lu kB\n", memlimit - memusage);
+			snprintf(lbuf, 100, "MemAvailable:   %8lu kB\n", memlimit - memusage + cached);
 			printme = lbuf;
 		} else if (startswith(line, "SwapTotal:") && memswlimit > 0) {
 			sscanf(line+sizeof("SwapTotal:")-1, "%lu", &hostswtotal);


asokoloski:
Forgot to add sign-off line.  I'll resubmit.

PR #237, opened 2018-05-02
loadavg

zhang2639:
This patch provides an interface for loadavg and consists of the load daemon, the hash table, and an interface of reading loadavg. Container can get its real loadavg by using it.

Hash table stores average load information for all containers.

The load daemon finds the container’s process pid through cgroup file, then gets all threads’ status by reading proc system. Finally it calculates the average load per container and refreshes them just as the Linux kernel did. See:
https://github.com/torvalds/linux/blob/master/kernel/sched/loadavg.c
(calculate the average load function)

Applications in the container can get the container’s real loadavg by using the interface. It should be noted that container must read the interface once when it is created in order to create and store the hash node of this container.

This function is not turned on by default. You can use it by using parameter __-l__. See:
./lxcfs -s -d -l -o allow_other /var/lib/lxcfs

Signed-off-by: zhang2639 zhangshiqiang@hust.edu.cn
Signed-off-by: yuwang yuwang@linux.alibaba.com

### DIFF
--- a/bindings.c
+++ b/bindings.c
@@ -66,6 +66,7 @@ enum {
 	LXC_TYPE_PROC_STAT,
 	LXC_TYPE_PROC_DISKSTATS,
 	LXC_TYPE_PROC_SWAPS,
+	LXC_TYPE_PROC_LOADAVG,
 };
 
 struct file_info {
@@ -4128,7 +4129,8 @@ int proc_getattr(const char *path, struct stat *sb)
 			strcmp(path, "/proc/uptime") == 0 ||
 			strcmp(path, "/proc/stat") == 0 ||
 			strcmp(path, "/proc/diskstats") == 0 ||
-			strcmp(path, "/proc/swaps") == 0) {
+			strcmp(path, "/proc/swaps") == 0 ||
+			strcmp(path, "/proc/loadavg") == 0) {
 		sb->st_size = 0;
 		sb->st_mode = S_IFREG | 00444;
 		sb->st_nlink = 1;
@@ -4148,7 +4150,8 @@ int proc_readdir(const char *path, void *buf, fuse_fill_dir_t filler, off_t offs
 	    filler(buf, "stat", NULL, 0) != 0 ||
 	    filler(buf, "uptime", NULL, 0) != 0 ||
 	    filler(buf, "diskstats", NULL, 0) != 0 ||
-	    filler(buf, "swaps", NULL, 0) != 0)
+	    filler(buf, "swaps", NULL, 0) != 0   ||
+	    filler(buf, "loadavg", NULL, 0) != 0)
 		return -EINVAL;
 	return 0;
 }
@@ -4170,6 +4173,8 @@ int proc_open(const char *path, struct fuse_file_info *fi)
 		type = LXC_TYPE_PROC_DISKSTATS;
 	else if (strcmp(path, "/proc/swaps") == 0)
 		type = LXC_TYPE_PROC_SWAPS;
+	else if (strcmp(path, "/proc/loadavg") == 0)
+		type = LXC_TYPE_PROC_LOADAVG;
 	if (type == -1)
 		return -ENOENT;
 
@@ -4227,6 +4232,8 @@ int proc_read(const char *path, char *buf, size_t size, off_t offset,
 		return proc_diskstats_read(buf, size, offset, fi);
 	case LXC_TYPE_PROC_SWAPS:
 		return proc_swaps_read(buf, size, offset, fi);
+	case LXC_TYPE_PROC_LOADAVG:
+		return proc_loadavg_read(buf, size, offset, fi);
 	default:
 		return -EINVAL;
 	}
--- a/bindings.c
+++ b/bindings.c
@@ -82,6 +82,21 @@ struct file_info {
 
 /* The function of hash table.*/
 #define LOAD_SIZE 100 /*the size of hash_table */
+#define FLUSH_TIME 5  /*the flush rate */
+#define DEPTH_DIR 3   /*the depth of per cgroup */
+/* The function of calculate loadavg .*/
+#define FSHIFT		11		/* nr of bits of precision */
+#define FIXED_1		(1<<FSHIFT)	/* 1.0 as fixed-point */
+#define EXP_1		1884		/* 1/exp(5sec/1min) as fixed-point */
+#define EXP_5		2014		/* 1/exp(5sec/5min) */
+#define EXP_15		2037		/* 1/exp(5sec/15min) */
+#define LOAD_INT(x) ((x) >> FSHIFT)
+#define LOAD_FRAC(x) LOAD_INT(((x) & (FIXED_1-1)) * 100)
+/* 
+ * This parameter is used for proc_loadavg_read().
+ * 1 means use loadavg, 0 means not use.
+ */
+static int loadavg = 0;
 static int calc_hash(char *name)
 {
 	unsigned int hash = 0;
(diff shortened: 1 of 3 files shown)


bindings.c
```diff
@@ -4092,6 +4292,363 @@ static int proc_swaps_read(char *buf, size_t size, off_t offset,
...
+	struct dirent *file;
+	FILE *f = NULL;
+	size_t linelen = 0;
+	char *line = NULL;
+	int pd;
+	char *path_dir, *path;
+	char **pid;
+
+	/* path = dpath + "/cgroup.procs" + /0 */
+	path = (char *)malloc(strlen(dpath) + 20);
```
brauner: Please don't cast the result of `malloc()`.

bindings.c
```diff
@@ -4092,6 +4292,363 @@ static int proc_swaps_read(char *buf, size_t size, off_t offset,
...
+	}
+
+	strcpy(path, dpath);
+	fd = openat(cfd, path, O_RDONLY);
+	if (fd < 0)
+		goto out;
+
+	dir = fdopendir(fd);
+	if (dir == NULL)
+		goto out;
```
brauner: If `fdopendir()` fails you still need to cleanup the file descriptor `fd`. Currently, you're leaking it below.

bindings.c
```diff
@@ -4092,6 +4292,363 @@ static int proc_swaps_read(char *buf, size_t size, off_t offset,
...
+
+	while (((file = readdir(dir)) != NULL) && depth > 0) {
+		if (strncmp(file->d_name, ".", 1) == 0)
+			continue;
```
brauner: Don't you want also continue when `..`?:
```
		if (strcmp(file>d_name, ".") == 0)
			continue;

		if (strcmp(file>d_name, "..") == 0)
			continue;
```

bindings.c
```diff
@@ -4092,6 +4292,363 @@ static int proc_swaps_read(char *buf, size_t size, off_t offset,
...
+		if (file->d_type == DT_DIR) {
+			/* path + '/' + d_name +/0 */
+			path_dir = (char *)malloc(strlen(path) + 2 + sizeof(file->d_name));
```
brauner: Please don't cast the result of `malloc()`.

bindings.c
```diff
@@ -4092,6 +4292,363 @@ static int proc_swaps_read(char *buf, size_t size, off_t offset,
...
+		goto out;
+
+	strcat(path, "/cgroup.procs");
+	fd = openat(cfd, path, O_RDONLY);
+	if (fd < 0)
+		goto out;
+
+	f = fdopen(fd, "r");
+	if (!f)
+		goto out;
```
brauner: You're leaking `fd` if `fdopen()` fails.

bindings.c
```diff
@@ -4092,6 +4292,363 @@ static int proc_swaps_read(char *buf, size_t size, off_t offset,
...
+
+	while (getline(&line, &linelen, f) != -1) {
+		pid = (char **)realloc(*pid_buf, sizeof(char *) * (sum + 1));
```
brauner: Please don't cast the result of `realloc()`.

bindings.c
```diff
@@ -4092,6 +4292,363 @@ static int proc_swaps_read(char *buf, size_t size, off_t offset,
...
+		if (pid == NULL) {
+			lxcfs_error("%s\n", "realloc error!");
+			for (; sum > 0; sum--)
+				free((*pid_buf)[sum - 1]);
+			sum = -1;
+			goto out;
+		}
+		*pid_buf = pid;
+		do {
+			*(*pid_buf + sum) = (char *)malloc(strlen(line) + 3);
```
brauner: Please don't cast the result of `malloc()`.

bindings.c
```diff
@@ -4092,6 +4292,363 @@ static int proc_swaps_read(char *buf, size_t size, off_t offset,
...
+	char **idbuf;
+	char proc_path[256];
+	int i, ret, run_pid = 0, total_pid = 0, last_pid = 0;
+	char *line = NULL;
+	size_t linelen = 0;
+	int sum, length;
+	DIR *dp;
+	struct dirent *file;
+
+	idbuf = (char **)malloc(sizeof(char *));
```
brauner: Please don't cast the result of `malloc()`.

bindings.c
```diff
@@ -4092,6 +4292,363 @@ static int proc_swaps_read(char *buf, size_t size, off_t offset,
...
+	if (sum == 0 || sum == -1)
+		goto out;
+
+	for (i = 0; i < sum; i++) {
+		/*clean up '\n' */
+		length = strlen(idbuf[i])-1;
+		idbuf[i][length] = '\0';
+		do {
+			ret = snprintf(proc_path, 256, "/proc/%s/task", idbuf[i]);
+		} while (ret < 0);
```
brauner: Hmm, why? That looks pretty dangerous to be honest. `snprintf()` isn't guaranteed to allocate any memory internally so the OOM kill won't kill it which means that this could spin indefinitely. If the cause `snprintf()` error is e.g. `EIO` I don't see who calling this multiple times would fix anything. I think you should just handle the error once.

bindings.c
```diff
@@ -4092,6 +4292,363 @@ static int proc_swaps_read(char *buf, size_t size, off_t offset,
...
+
+		dp = opendir(proc_path);
+		if (!dp) {
+			lxcfs_error("%s\n", "Open proc_path failed in refresh_load.");
+			continue;
+		}
+		while ((file = readdir(dp)) != NULL) {
+			if (strncmp(file->d_name, ".", 1) == 0)
+				continue;
```
brauner: Same comment as above. You might want to continue on `..` too.

bindings.c
```diff
@@ -4092,6 +4292,363 @@ static int proc_swaps_read(char *buf, size_t size, off_t offset,
...
+			total_pid++;
+			/* We make the biggest pid become last_pid.*/
+			ret = atof(file->d_name);
+			last_pid = (ret > last_pid) ? ret : last_pid;
+
+			do {
+				ret = snprintf(proc_path, 256, "/proc/%s/task/%s/status", idbuf[i], file->d_name);
+			} while (ret < 0);
```
brauner: Same comment as above:
> Hmm, why? That looks pretty dangerous to be honest. snprintf() isn't guaranteed to allocate any memory internally so the OOM kill won't kill it which means that this could spin indefinitely. If the cause snprintf() error is e.g. EIO I don't see who calling this multiple times would fix anything. I think you should just handle the error once.

bindings.c
```diff
@@ -4092,6 +4292,363 @@ static int proc_swaps_read(char *buf, size_t size, off_t offset,
...
+				pthread_mutex_unlock(&load_hash[i].lock);
+				continue;
+			}
+			f = load_hash[i].next;
+			first_node = 1;
+			while (f) {
+				length = strlen(f->cg) + 2;
+				do {
+					/* strlen(f->cg) + '.' or '' + \0 */
+					path = (char *)malloc(length);
```
brauner: Please don't cast the result of `malloc()`.
Can you explain why you do a looping `malloc()` here? I mean it's fine as this will be killed by the OOM kill but I'd like to know why here and not in other places.

bindings.c
```diff
@@ -4092,6 +4292,363 @@ static int proc_swaps_read(char *buf, size_t size, off_t offset,
...
+				if (ret < 0 || ret > length - 1) {
+					/* snprintf failed, ignore the node.*/
+					lxcfs_error("Refresh node %s failed for snprintf().\n", f->cg);
+					goto out;
+				}
+				sum = refresh_load(f, path);
+				if (sum == 0) {
+					f = del_node(f, i);
+				} else {
+out:				f = f->next;
```
brauner: Wrong indentation.

bindings.c
```diff
@@ -4092,6 +4292,363 @@ static int proc_swaps_read(char *buf, size_t size, off_t offset,
...
+	if (n == NULL) {
+		if (!find_mounted_controller("cpu", &cfd)) {
+			/*
+			 * In locate_node() above, pthread_rwlock_unlock() isn't used
+			 * because delete is not allowed before read has ended.
+			 */
+			pthread_rwlock_unlock(&load_hash[hash].rdlock);
+			return 0;
+		}
+		n = (struct load_node *)malloc(sizeof(struct load_node));
```
brauner: Please don't cast the result of `malloc()`.

bindings.c
```diff
@@ -4092,6 +4292,363 @@ static int proc_swaps_read(char *buf, size_t size, off_t offset,
...
+		if (!n) {
+			lxcfs_error("Create node %s failed for malloc().\n", cg);
+			pthread_rwlock_unlock(&load_hash[hash].rdlock);
+			return read_file("/proc/loadavg", buf, size, d);
+		}
+		do {
+			n->cg = (char *)malloc(strlen(cg)+1);
```
brauner: Please don't cast the result of `malloc()`.

brauner: Same question as further above why loop around `malloc()` here and not in other places?

lxcfs.c
```diff
@@ -859,6 +864,9 @@ int main(int argc, char *argv[])
 	swallow_arg(&argc, argv, "-s");
 	swallow_arg(&argc, argv, "-f");
 	debug = swallow_arg(&argc, argv, "-d");
+	if (swallow_arg(&argc, argv, "-l")) {
```
brauner: Yes, I like that it's opt-in. Cool!

lxcfs.c
```diff
@@ -936,6 +936,28 @@ int main(int argc, char *argv[])
...
+	if (load_use == true) {
+		s = pthread_cancel(pid);
+		if (s == 0) {
+			s = pthread_join(pid, NULL); /* Make sure sub thread has been canceled. */
+			if (s != 0) {
+				lxcfs_error("%s\n", "load_free error!");
+				goto out;
+			}
+			dlerror();    /* Clear any existing error */
+			void (*load_free)(void);
```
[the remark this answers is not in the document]
zhang2639: OK.
At first, I think load_free might not be used, so that the previous statement would be redundant.
I will rewrite it. @hallyn

bindings.c
```diff
@@ -4244,6 +4259,363 @@ static int proc_swaps_read(char *buf, size_t size, off_t offset,
...
+					goto out;
+				}
+				sum = refresh_load(f, path);
+				if (sum == 0) {
+					f = del_node(f, i);
+				} else {
+out:				f = f->next;
+				}
+				free(path);
+				/* load_hash[i].lock locks only on the first node.*/
```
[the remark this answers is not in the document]
zhang2639: Insert_node() inserts the node into the head of the hash chain. So refreshing(deleting) the hash chain's header and insert should be mutually exclusive.
To this *A slow load_begin() could, while on first_node == 0, race with a load_free() right?*, I use load_free() after using _pthread_cancel_ and _pthread_join_. So there should be no race here.

bindings.c
zhang2639: Sorry, my English is a bit poor. I didn’t understand the advice you gave me before about __cast__.
Now I understand.  Don't cast the result of __malloc()__ is not the same thing I learned in the course and is very cool. I will rewrite them. 
Thank you! @brauner

bindings.c
zhang2639: Yes, you're right.
I will rewrite them.

bindings.c
zhang2639: Unbelievable, I always thought I had already written these about __..__ .
I will add them.

bindings.c
zhang2639: OK.

bindings.c
zhang2639: OK, I will rewrite them.

bindings.c
zhang2639: OK.

bindings.c
zhang2639: OK.

bindings.c
zhang2639: OK.

bindings.c
zhang2639: I declared __proc_path[256]__ in the beginning. In this case, can proc_path not get 256 bytes of space? I am puzzled. @brauner

bindings.c
zhang2639: At first I want to use loops in all malloc() situations. But I don’t know whether it is safe. 
It's hard to say why I do a looping malloc() here.
I will use loops in all malloc() situations.
Thank you!

bindings.c
zhang2639: The format on the local is correct, but I don't know why it was wrong after pushing.
I will try to solve it.

bindings.c
zhang2639: OK.

zhang2639:
Thank for reviewing them.
I will rewrite them later.

bindings.c
brauner: Whether to cast the result of `malloc()` or not is a worthwhile discussion to have. :) We have never cast the result of `malloc()` in all our projects, the Linux kernel - whose coding style we usually follow - doesn't either. In case you're interested in pro and con arguments for casting:

- [contra](https://stackoverflow.com/questions/1565496/specifically-whats-dangerous-about-casting-the-result-of-malloc/)
- [pro](https://wiki.sei.cmu.edu/confluence/display/c/MEM02-C.+Immediately+cast+the+result+of+a+memory+allocation+function+call+into+a+pointer+to+the+allocated+type)

bindings.c
brauner: Make sure that you only close the `fd` via `close(fd)` when `fdopendir()` fails. After `fdopendir()` has returned successfully you can only close `dir` via `closedir(dir)`.

bindings.c
brauner: So, there are two problems here:
1. `snprintf()` cal fail for reasons other than too small a buffer size. One of the ways I gave as an example is `EIO`.
2. You can't loop around `snprint()` as nothing guarantees that this loop ends. It could go on forever.

bindings.c
zhang2639: It sounds good and is worth learning.

bindings.c
zhang2639: 👍 
Got it！

bindings.c
zhang2639: 👍 
OK. It's dangerous. 
I will rewrite them.

hallyn:
Thanks - with the feedback addressed, +1 from me (though since I gather malloc will be changing i'll take another look at that when you're ready)

zhang2639:
I have rewritten the loadavg. Please review it.
Thank you very much！
@hallyn @brauner

brauner:
I'm going to pull the branch and do some tests before merging.

brauner:
Seems to work fine.

lxcfs.c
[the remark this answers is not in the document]
haosdent: @zhang2639 have you consider to submit a merge request to fix /sys/devices/system/cpu/online as well?

PR #241, opened 2018-06-13
bindings: fix memory leak in proc_loadavg_read()

aither64:
In `proc_loadavg_read()`, `cg` is allocated by`get_pid_cgroup()`, but it's not freed before returning.


brauner:
Thanks!

PR #242, opened 2018-06-13
loadavg: restart thread on library reload

aither64:
Per-container loadavg tracking is not working after library reload using `SIGUSR1`. When reloading the library in `do_reload()`, the thread that is tracking per-container loadavgs is left running. As I understand it, it continues to operate on those static variables from the original library. loadavg is not set up using the new library and so `proc_read_loadavg()` starts to return the host's `/proc/loadavg`, as if it was disabled.

This patch stops the loadavg thread before closing the original library and restarts it when the new library is loaded.

### DIFF
--- a/lxcfs.c
+++ b/lxcfs.c
@@ -69,6 +69,46 @@ static void users_unlock(void)
 	unlock_mutex(&user_count_mutex);
 }
 
+static pthread_t loadavg_pid = 0;
+
+/* Returns zero on success */
+static int start_loadavg(void) {
+	char *error;
+	pthread_t (*load_daemon)(int);
+
+	dlerror();    /* Clear any existing error */
+
+	load_daemon = (pthread_t (*)(int)) dlsym(dlopen_handle, "load_daemon");
+	error = dlerror();
+	if (error != NULL) {
+		lxcfs_error("load_daemon fails:%s\n", error);
+		return -1;
+	}
+	loadavg_pid = load_daemon(1);
+	if (loadavg_pid == 0)
+		return -1;
+
+	return 0;
+}
+
+/* Returns zero on success */
+static int stop_loadavg(void) {
+	char *error;
+	int (*stop_load_daemon)(pthread_t);
+
+	stop_load_daemon = (int (*)(pthread_t)) dlsym(dlopen_handle, "stop_load_daemon");
+	error = dlerror();
+	if (error != NULL) {
+		lxcfs_error("stop_load_daemon error: %s\n", error);
+		return -1;
+	}
+
+	if (stop_load_daemon(loadavg_pid) != 0)
+		return -1;
+
+	return 0;
+}
+
 static volatile sig_atomic_t need_reload;
 
 /* do_reload - reload the dynamic library.  Done under
@@ -76,6 +116,10 @@ static volatile sig_atomic_t need_reload;
 static void do_reload(void)
 {
 	char lxcfs_lib_path[PATH_MAX];
+
+	if (loadavg_pid > 0)
+		stop_loadavg();
+
 	if (dlopen_handle) {
 		lxcfs_debug("%s\n", "Closing liblxcfs.so handle.");
 		dlclose(dlopen_handle);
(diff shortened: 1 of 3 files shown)


lxcfs.c
```diff
@@ -922,42 +975,13 @@ int main(int argc, char *argv[])
 	if ((pidfd = set_pidfile(pidfile)) < 0)
 		goto out;
 
-	if (load_use == true) {
```
brauner: That should just be `if (load_use)`. :)

lxcfs.c
```diff
@@ -922,42 +975,13 @@ int main(int argc, char *argv[])
...
-		dlerror();    /* Clear any existing error */
+	if (load_use == true && start_loadavg() != 0)
```
brauner: Same, that should just be `if (load_use)`. :)

lxcfs.c
```diff
@@ -922,42 +975,13 @@ int main(int argc, char *argv[])
...
-			if (error != NULL) {
-				lxcfs_error("load_free error: %s\n", error);
-				goto out;
-			}
-			load_free();
-		} else {
-			lxcfs_error("%s\n", "load_free error!");
-		}
-	}
+	if (load_use == true)
```
brauner: That should just be `if (load_use)`. :)

brauner:
Did you test this?

brauner:
And thanks!. :)

aither64:
I'm glad that you didn't merge it right away, since I've had to rework it a bit. It worked alright *most of the time*, except when `do_reload()` was called while the loadavg thread was refreshing those averages. `pthread_cancel()` killed the thread and some mutexes apparently remained locked, which blocked `lock_free()` forever.

I've made it so that the thread is stopped safely from the inside after releasing mutexes, not cancelled from the outside. It now takes several seconds to stop the thread though, I'm not sure how to improve that.

If it's ok now, let's wait a while before merging. I can try it out on our staging node with some load on it, where I noticed the issue in the first place.

brauner:
@aither64, cool thanks!

brauner:
Ok, @stgraber, @hallyn, please don't merge just yet. :)

brauner:
> It now takes several seconds to stop the thread though, I'm not sure how to improve that.

I suspect this happens because the threads can't be cancelled asynchronously and so have to wait for a cancellation point. You can try and experiment with `PTHREAD_CANCEL_ASYNCHRONOUS` and see if that helps. But you need to bring forward a good argument that it is safe to use in our case here. :)
Also, we should CC the original authors of the patch too. :) //cc @zhang2639 can you give this a look too, please. :)

aither64:
> I suspect this happens because the threads can't be cancelled asynchronously and so have to wait for a cancellation point.

No, the delay is there because the loadavg thread sleeps for about 5 seconds between refreshes, see the loop in `load_begin()`. The main thread now signals the loadavg thread to exit by setting `loadavg_stop = 1`, then waits for it to wake, exit and join, hence the slowness. I've removed `pthread_cancel()` to avoid those cancellation points, because I suspect it can cause deadlocks.

One sure way to trigger a deadlock with `pthread_cancel()` that I found is to put some `sleep()` into the loop in `load_begin()`, between mutex lock&unlock, wait for it to execute and then send `SIGUSR1`. The [original patch](https://github.com/aither64/lxcfs/commit/5235d372dd58e064deb2e95dc2c690c31774f9e2) failed this test, the current one works fine. AFAIK, there are other possible cancellation points between mutex lock&unlock (e.g. in `refresh_load()`), so given the right circumstances, the thread can be cancelled before it does the unlock. It seems better to me to stop the thread cleanly instead of killing it.

Anyway, I've tested it and it works fine for us.

> Also, we should CC the original authors of the patch too. :) //cc @zhang2639 can you give this a look too, please. :)

Yeah, should have thought about that, apologies :)

brauner:
Ok, so we're good to merge then. :)

zhang2639:
Sorry, I just saw it. I did not consider `do_reload()` before. Your work is admirable. Thank you！

Issue #180, opened 2017-03-14
CPU usage is shared among VMs


servereasy-saverio:
Hi,

Inside the VMs, _top_ shows the CPU usage of the whole node.

hallyn:
Do you have your containers configured to isolate cpus or cpushare?  Can you show your container configuration and precisely which fields are not being properly virtualized?

servereasy:
cpushare:

...
limits.cpu: "8"
...

For example, if inside "VM A" I produce 100% CPU usage, inside "VM B" I see 100% CPU usage (if the assigned cores are the same, of course)

kovanorby:
Hello!

I'm usign proxmox, but on the lxc vps i only see the load of the physical cpus... It's not good this way

hallyn:
If the containers are sharing cpus then indeed they will see each others' loads.

If someone would like to work an a good solution to this, please feel free (and post here your plans and progress and questions) - we'll leave this issue open to track that.  If I find some time to work on it I'll do so.

As a workaround you can dedicate cpus to specific containers.  Of course that's sub-optimal if you have 4 cpus and 1k containers.

servereasy-saverio:
Hello,

Is there any plan about this?

aither64:
I've tried to implement this feature, but I'm not sure if there is actually a way to do this properly without the kernel providing more information. I've made it so that LXCFS reads CPU usage from the container's cpuacct cgroup and then uses those values to override `/proc/stat` given to the container. The issue is that the cgroup is accounting only usage in `user` and `system` modes. I was able to calculate idle time based on values from `/proc/stat`, but it may not be very accurate, because accounted usage by `/proc/stat` and the cgroup differs a bit... I've no idea what to do about other fields in `/proc/stat`, like `nice`, `iowait` or interrupts, so those fields are always set to `0`. Here's the patch: https://github.com/aither64/lxcfs/tree/independent_cpu_usage

@hallyn, what do you think about it, is there a better way?

N-Storm:
Maybe take a look at OpenVZ code? This was implemented there for ages. As for LXC this issue was raised so many times for years but nobody cares. Like #40 was closed without any comments.

hallyn:
@N-Storm as we said, this requires kernel support.  That's how OpenVZ did it.  lxc aims to work purely with upstream kernel (or at least patches on their way upstream).

[...]

PR #244, opened 2018-06-15
Per-container CPU usage in /proc/stat

aither64:
As it is, CPU usage stats in `/proc/stat` are shared with the host and all containers (issue #180). I'm trying to find a way for each container to see only its own CPU usage.

With this patch, `proc_stat_read()` reads per-container CPU usage from the *cpuacct* cgroup, `cpuacct.usage_all` to be exact. However, this file contains only usage in `user` and `system` modes, other fields from `/proc/stat` are not present. Idle time can be calculated, but as for other values like `nice` or `iowait`, they're all reported as zeroes in the resulting `/proc/stat` file. I don't think this can be solved without the kernel providing more information.

*cpuacct* cgroup v1 has to be present, otherwise `proc_stat_read()` will return an error. This could be changed if needed, I don't know if LXCFS is supposed to work without this cgroup somewhere.

We've been using it for several weeks on a staging node with [vpsAdminOS](https://github.com/vpsfreecz/vpsadminos) and it seems to work as expected.

### DIFF
--- a/bindings.c
+++ b/bindings.c
@@ -80,6 +80,11 @@ struct file_info {
 	int cached;
 };
 
+struct cpuacct_usage {
+	uint64_t user;
+	uint64_t system;
+};
+
 /* The function of hash table.*/
 #define LOAD_SIZE 100 /*the size of hash_table */
 #define FLUSH_TIME 5  /*the flush rate */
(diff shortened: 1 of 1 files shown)


brauner:
How does this deal with the case where the `cpuacct` controller is missing. It seems to me that we should just return the host values in this case and not fail/error.

aither64:
Alright, it's going to require more code and testing though, I haven't thought about it before. I will let you know when I have something, next week or so.

aither64:
That is, if otherwise this patch is fine with you -- those zeroes instead of many fields in `/proc/stat`, I can't do anything about that afaik.

brauner:
@aither64, it might be acceptable. I have no immediate objections to it. @hallyn will need to yay or nay as well though. :)

hallyn:
zeroes where we can't get real values is fine, I think.  (An error when cpuacct.v1 is not available is not).

thanks.

aither64:
Thank you for the info.

I've updated the patch. It now reads from `cpuacct` only if it is available, otherwise it falls back to the host's `/proc/stat`, just as it worked before. I've tested it in my environment and it looks good. It would be great if someone else could test this as well.

One additional imperfection that I've noticed is that CPU idle time is still leaking from the host, i.e. when you start a new container, `user` and `system` counters start from zero, but `idle` is based on the host's current value. I don't think it's a big issue, since I imagine tools working on usage data from `/proc/stat` are looking for relative changes of those counters and not the absolute values.

@brauner, @hallyn, please check it out. Thanks!

bindings.c
```diff
@@ -3794,6 +3799,90 @@ static uint64_t get_reaper_age(pid_t pid)
...
+	struct cpuacct_usage *cpu_usage;
+	int rv = 0, i, j, ret, read_pos = 0, read_cnt;
+	int cg_cpu;
+	uint64_t cg_user, cg_system;
+	int64_t ticks_per_sec;
+	char *usage_str = NULL;
+
+	ticks_per_sec = sysconf(_SC_CLK_TCK);
+
+	if (ticks_per_sec < 0 && errno == EINVAL) {
```
brauner: Nit: unnecessary new line :)

bindings.c
```diff
@@ -3794,6 +3799,90 @@ static uint64_t get_reaper_age(pid_t pid)
...
 
+/*
+ * Returns 0 on success.
+ * It is the caller's responsibility to free `return_usage`, unless this
+ * function returns an error.
+ */
+static int read_cpuacct_usage_all(char *cg, char *cpuset, struct cpuacct_usage **return_usage)
+{
+
+	int cpucount = get_nprocs();
```
brauner: `get_nprocs()` always returns a positive value? If not the `malloc()` below will be might funny...

bindings.c
```diff
@@ -3794,6 +3799,90 @@ static uint64_t get_reaper_age(pid_t pid)
...
+	struct cpuacct_usage *cpu_usage;
+	int rv = 0, i, j, ret, read_pos = 0, read_cnt;
```
brauner: Nit: If you do `int rv = -1` here you can spare all those `rv = -1` in your error paths and simply return and also remove `{`, `}`. :)

bindings.c
```diff
@@ -3794,6 +3799,90 @@ static uint64_t get_reaper_age(pid_t pid)
...
+	read_pos += read_cnt;
+
+	for (i = 0, j = 0; i < cpucount; i++) {
+		ret = sscanf(usage_str + read_pos, "%d %lu %lu\n%n", &cg_cpu, &cg_user,
+				&cg_system, &read_cnt);
+
+		if (ret == EOF) {
+			break;
+
+		} else if (ret != 3) {
```
brauner: Nit: There's no need for the `if else` ladder here. You can just do:
```
if (ret == EOF)
        break;

if (ret != 3) {
.
.
.
```

bindings.c
aither64: Manpage says: *The function get_nprocs() returns the number of processors currently available in the  system.* So I expect it should be always at least `1`.

bindings.c
aither64: That's true, but I prefer to make the return value explicit, e.g. someone can change the default `rv` in the future and it won't break.

bindings.c
aither64: Fixed

bindings.c
aither64: Fixed

Issue #222, opened 2017-11-05
Wrong RAM usage report in htop / ubuntu 16.04


lug-gh:
Hi, on containers with ubuntu 16.04 i get wrong ram usage in htop

lxc-ls --version
2.1.0

lxcfs --version
2.0.7

![yos](https://lickmynip.pl/share/i/6ny63orn77.png)

free -m
              total        used        free      shared  buff/cache   available
Mem:            512         204         292         956          14         292
Swap:             0           0           0

ram usage shown in proxmox: 142 mb

so everything shows different values, which one to trust?

brauner:
This has been changed by @Blub and there's really not nice way to do this since memory info with cgroup v1 is not very nice.

lug-gh:
Ehm I don't really understand what you're trying to tell me :(

brauner:
Oh, I mixed your issue with another one it seems. Ok, so is this reproducible with current `LXCFS` aka `2.0.8` and `3.0.0`?

lug-gh:
lxc-ls 3.0.0 and lxcfs 3.0.0

![free -m](https://lickmynip.pl/share/i/ojyi45a81p.png)
![htop](https://lickmynip.pl/share/i/eqzkqudpfa.png)

aither64:
I've noticed wrong memory usage info in htop as well. In our case, the issue is that in `/proc/meminfo`, LXCFS is handling `Cached` and `SReclaimable`, but `Shmem` is passed as-is from the host. @lug-gh, do you have high `Shmem` in `/proc/meminfo` as well? It should be the difference between the reported memory usage in htop and free.

The way htop calculates memory usage is this (see [here](https://github.com/hishamhm/htop/blob/666e1e76b39ee66a38f5fb620d22b23f36859eca/linux/LinuxProcessList.c#L960) and [here](https://github.com/hishamhm/htop/blob/666e1e76b39ee66a38f5fb620d22b23f36859eca/linux/Platform.c#L203)):
```
cachedMem = Cached + SReclaimable - Shmem
usedMem = MemTotal - MemFree - Buffers - cachedMem
```
Since `Cached` is low, `SReclaimable` is always zero and `Shmem` is taken from the host and can get quite large, `cachedMem` becomes a negative number and actually increases `usedMem`.

[...]

PR #248, opened 2018-07-04
meminfo: read shmem from memory cgroup

aither64:
In `/proc/meminfo`, `Shmem` is passed as-is from the host while other values aren't. This causes `htop` to show incorrect memory usage, as described in #222. This patch makes it so that:

- `Shmem` is read from `memory.stat`. If it's not found, it is set to `0`.
- `ShmemHugePages` and `ShmemPmdPages` are always set to `0`, as I haven't found a cgroup counterpart. These values aren't related to the problem with `htop`.


brauner:
lgtm

Issue #239, opened 2018-05-21
Does lxcfs have plans to support cpu-shares and cpu-quota?


agile6v:
Hello.

Lxcfs currently does not support the CPU views requested by the cpu-shares and cpu-quota in the container. Is there plan support?

Regards.
agile6v

stgraber:
Not sure how that'd get rendered in cpuinfo

agile6v:
@stgraber I understand your doubts. If I run a container with `--cpuset-cpus` parameter, then the CPU view seen in the container is bound by this container, but the CPU resource is used by this container alone? Obviously not. If so, I think a container with `--cpu-shares` & `--cpu-quota` parameter can randomly display corresponding number of CPU in the container.

aither64:
If I understand it correctly that you'd like to reduce the number of visible CPUs based on `cpu.shares` or `cpu.cfs_{quota,period}_us`, I'd be interested in it as well.

@stgraber, could you please elaborate on what problems do you see with `/proc/cpuinfo`? I think it could calculate the number of available CPUs based on shares/quotas/cpuset, then copy only as many CPUs from the host's `/proc/cpuinfo` and ignore the rest. It works like this even now, just with cpuset, doesn't it?

`/proc/stat` would be more complicated, I think. Since a container can utilize more CPUs than we might want to be visible, the usage counters from hidden CPUs would somehow need to be added to the visible CPUs. It should be possible with #244 merged in.

I'm probably going to be experimenting with this.

Zexi:
@aither64 Hi, I think this feature is a very useful, is there something new about your experimenting?

aither64:
Yes, I have it [implemented](https://github.com/lxc/lxcfs/compare/master...aither64:cpu-views), but I haven't gotten around to cleaning it up and making a PR yet. It works with CPU quotas set e.g. using `lxc.cgroup.cpu.cfs_{quota,period}_us`, CPU shares didn't make sense to me.

Zexi:
@aither64 Thanks for your reply. As expected, I tested this feature and it works. It would be nice if make your [implemented](https://github.com/lxc/lxcfs/compare/master...aither64:cpu-views) merged.

lijiaocn:
@agile6v HI,  I try to use `lxcfs` with cpu-view feature， It seems not work for me. I use docker:

`docker run -it --rm -m 256m   --cpu-quota 1000 \
[...]

PR #260, opened 2018-10-23
CPU views based on quotas

aither64:
Currently, containers can see all CPUs in `/proc/stat` except those excluded using cpuset. cpuset isn't very practical when you have a lot of containers with unpredictable load, so we're using CPU quotas to limit CPU usage, which let the containers utilize all CPUs within the quota as the scheduler sees fit. This patchset reduces the number of CPUs visible in `/proc/cpuinfo` and `/proc/stat` based on CPU quotas configured e.g. using `lxc.cgroup.cpu.cfs_{quota,period}_us`. If qoutas aren't set, it falls back to the current behaviour.

Since containers can utilize more CPUs then should be visible, LXCFS generates per-container CPU usage view. It has to remember what values were reported to containers, so that the usage counters are continuously increasing and to ensure that no CPU will be utilized for more than 100% between two consecutive reads of `/proc/stat`. This is done by introducing state. CPU usage of all containers is tracked in a hash table with linked lists, similar to the loadavg code. CPU usage counters from hidden CPUs are spread over the visible CPUs that are not already fully utilized within the time frame since the last read of `/proc/stat`, starting from the first CPU to the last. This means that the CPU usage as seen within containers does not reflect the real usage on the host system.

Like with `/proc/loadavg`, the CPU usage accounting for containers starts to work on the first read of `/proc/stat`. CPU views are used only if there is cpuacct cgroup, cpu cgroup and `cpu.cfs_{quota,period}_us` are set.

Discussed also in #239.


bindings.c
```diff
@@ -3989,6 +4078,420 @@ static int read_cpuacct_usage_all(char *cg, char *cpuset, struct cpuacct_usage *
...
+			   &softirq,
+			   &steal,
+			   &guest,
+			   &guest_nice);
+
+		if (ret != 10)
+			continue;
+
+		all_used = user + nice + system + iowait + irq + softirq + steal + guest + guest_nice;
+		cg_used = cg_cpu_usage[curcpu].user + cg_cpu_usage[curcpu].system;
```
hallyn: What makes it safe to not have an assert for curcpu being < sizeof(cg_cpu_usage)/sizeof(*cg_cpu_usage) ?

bindings.c
```diff
@@ -4140,10 +4213,14 @@ static struct cg_proc_stat *find_proc_stat_node(const char *cg)
...
 		if (strcmp(cg, node->cg) == 0)
-			return node;
+			goto out;
 	} while ((node = node->next));
 
-	return NULL;
+	node = NULL;
+
+out:
+	prune_proc_stat_history();
```
hallyn: I could imagine this going wrong based on timing.  You take the lock;  find the entry, make sure it's not stale;  then you drop the lock, call prune, which takes the lock and might drop this node if it by now has become stale;  finally you return this node unconditionally, but it may have been pruned, right?

aither64: Right, but I don't think that it can happen because of the way the function is used. Nodes are pruned when the underlying cgroup no longer exists, but since the call is made by a process from the cgroup, it has to exist and thus the node cannot be pruned.

aither64:
I've updated the code to address @hallyn's comments, except the pruning, see my comment above if it makes sense. Changes:

 - explicitly initialize struct members in `new_proc_stat_node()`
 - cpu_view -> cpuview
 - out of bounds checks for cpuacct usage data access as a new commit

PR #262, opened 2018-12-12
bindings: prevent NULL pointer dereference

brauner:
Fixes: https://bugs.launchpad.net/ubuntu/+source/lxcfs/+bug/1807628

### DIFF
--- a/bindings.c
+++ b/bindings.c
@@ -2136,7 +2136,7 @@ int cg_readdir(const char *path, void *buf, fuse_fill_dir_t filler, off_t offset
 		goto out;
 	}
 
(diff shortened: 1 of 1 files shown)


Blub:
As far as I can tell this can only be `NULL` when lxcfs encounteres a cgroup directory with no non-directory files present. I'm not sure how or why that would happen, but from the code's perspective the check makes sense (and we do it for the directory case below as well).

axinojolais:
Hi, do we have an idea hen will this be released ? Thanks

brauner:
@stgraber, any release planned?

PR #266, opened 2019-01-18
Adds mutex to do_release_file_info 

tomponline:
Adds mutex to do_release_file_info to avoid lxcfs crash when stopping multiple containers concurrently.

Fixes https://github.com/lxc/lxcfs/issues/252

### DIFF
--- a/bindings.c
+++ b/bindings.c
@@ -2169,11 +2169,16 @@ out:
 	return ret;
 }
 
+static pthread_mutex_t do_release_file_info_lock = PTHREAD_MUTEX_INITIALIZER;
+
 static void do_release_file_info(struct fuse_file_info *fi)
 {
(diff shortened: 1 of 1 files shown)


tomponline:
thanks @brauner :)

brauner:
On Fri, Jan 18, 2019 at 07:29:42AM -0800, Wolfgang Bumiller wrote:
> IMO this would indicate a reference counting issue in libfuse. From lxcfs' perspective it doesn't make sense to get simultaneous release callbacks for the same object, as you cannot lock efficiently then. This is a global lock for releases, which seems rather expensive. However, moving the mutex into the file-info would not be sound, so I don't have any other solution for this for now.

For sure. This is really just a stop-gap.

> 
> Thanks for the patch.

PR #267, opened 2019-01-22
 Revert "bindings: protect do_release_file_info() with mutex"

tomponline:
@brauner I'm still experiencing some lxcfs hangs with this change, so probably best to revert it until I can investigate further.

Thanks

### DIFF
--- a/bindings.c
+++ b/bindings.c
@@ -2169,16 +2169,11 @@ out:
 	return ret;
 }
 
-static pthread_mutex_t do_release_file_info_lock = PTHREAD_MUTEX_INITIALIZER;
-
 static void do_release_file_info(struct fuse_file_info *fi)
 {
-	lock_mutex(&do_release_file_info_lock);
-
 	struct file_info *f = (struct file_info *)fi->fh;
 
 	if (!f)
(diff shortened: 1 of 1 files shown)


tomponline:
@Blub @brauner It seems that there is already a mutex protecting the higher functions that call this function anyway:

https://github.com/lxc/lxcfs/blob/8437cc6cb6b424cdb81db35915b7914ef934e736/lxcfs.c#L587-L589

tomponline:
@brauner its quite strange, adding some error logging around the locks seems to fix it, but taking the debug lines starts to introduce hanging again (like its getting stuck on a deadlock).

tomponline:
The `up_users` function just does a temporary lock to incremnt/decrement a counter, doesn't actually lock the actions that happen in the middle.

Blub:
> adding some error logging around the locks seems to fix it

Races are tricky that way. Printing uses syscalls which introduces context switches between kernel and user space, which can affect the timing a *lot*. Any debug helpers you introduce will affect the timing in an unpredictable way. That's what makes race conditions so annoying to hunt.

PR #271, opened 2019-02-18
option to disable swap in meminfo

dmaj:
Signed-off-by: Dieter Maier <dma@web.de>

It concerns the combination of a system with activated swap and docker. When I use lxcfs to limit the memory resources of a container, I always have swap.. Which I don't want.

docker run -it -m 256m --memory-swap 256m \
      -v /var/lib/lxcfs/proc/cpuinfo:/proc/cpuinfo:rw \
      -v /var/lib/lxcfs/proc/diskstats:/proc/diskstats:rw \
      -v /var/lib/lxcfs/proc/meminfo:/proc/meminfo:rw \
      -v /var/lib/lxcfs/proc/stat:/proc/stat:rw \
      -v /var/lib/lxcfs/proc/swaps:/proc/swaps:rw \
      -v /var/lib/lxcfs/proc/uptime:/proc/uptime:rw \
      ubuntu:18.04 /bin/bash

That, too, is not satisfactory:
docker run -it -m 256m \
      -v /var/lib/lxcfs/proc/cpuinfo:/proc/cpuinfo:rw \
      ....

I can force with the parameter "-u" that the swap values in the "meminfo" are always 0.

To use this parameter I use "fuse_get_context()->private_data".

It is a very special use case, but I wanted to at least suggest it.

I haven't programmed a C for a long time, so please take a critical look at it.

It works very well with Ubuntu. Unfortunately not with Alpine yet. I tested it with "free".

I developed the customization to use several mongodb on one docker host, which unfortunately have swap enabled. This works very well.

### DIFF
--- a/bindings.c
+++ b/bindings.c
@@ -3437,6 +3437,7 @@ static int proc_meminfo_read(char *buf, size_t size, off_t offset,
 		struct fuse_file_info *fi)
 {
 	struct fuse_context *fc = fuse_get_context();
+	struct lxcfs_opts *opts = (struct lxcfs_opts *) fuse_get_context()->private_data;
 	struct file_info *d = (struct file_info *)fi->fh;
 	char *cg;
 	char *memusage_str = NULL, *memstat_str = NULL,
@@ -3517,17 +3518,23 @@ static int proc_meminfo_read(char *buf, size_t size, off_t offset,
 		} else if (startswith(line, "MemAvailable:")) {
 			snprintf(lbuf, 100, "MemAvailable:   %8lu kB\n", memlimit - memusage + cached);
 			printme = lbuf;
-		} else if (startswith(line, "SwapTotal:") && memswlimit > 0) {
+		} else if (startswith(line, "SwapTotal:") && memswlimit > 0 && opts->swap_off == false) {
 			sscanf(line+sizeof("SwapTotal:")-1, "%lu", &hostswtotal);
 			if (hostswtotal < memswlimit)
 				memswlimit = hostswtotal;
 			snprintf(lbuf, 100, "SwapTotal:      %8lu kB\n", memswlimit);
 			printme = lbuf;
-		} else if (startswith(line, "SwapFree:") && memswlimit > 0 && memswusage > 0) {
+		} else if (startswith(line, "SwapTotal:") && opts->swap_off == true) {
+			snprintf(lbuf, 100, "SwapTotal:      %8lu kB\n", 0UL);
+			printme = lbuf;
+		} else if (startswith(line, "SwapFree:") && memswlimit > 0 && memswusage > 0 && opts->swap_off == false) {
 			unsigned long swaptotal = memswlimit,
 					swapusage = memswusage - memusage,
 					swapfree = swapusage < swaptotal ? swaptotal - swapusage : 0;
 			snprintf(lbuf, 100, "SwapFree:       %8lu kB\n", swapfree);
+			printme = lbuf;
+		} else if (startswith(line, "SwapFree:") && opts->swap_off == true) {
+			snprintf(lbuf, 100, "SwapFree:       %8lu kB\n", 0UL);
 			printme = lbuf;
 		} else if (startswith(line, "Slab:")) {
 			snprintf(lbuf, 100, "Slab:        %8lu kB\n", 0UL);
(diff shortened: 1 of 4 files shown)


brauner:
I'm not against it. I'm off today so won't get to review it but will check in later.

bindings.c
```diff
@@ -3517,18 +3518,32 @@ static int proc_meminfo_read(char *buf, size_t size, off_t offset,
...
-		} else if (startswith(line, "SwapTotal:") && memswlimit > 0) {
-			sscanf(line+sizeof("SwapTotal:")-1, "%lu", &hostswtotal);
-			if (hostswtotal < memswlimit)
-				memswlimit = hostswtotal;
-			snprintf(lbuf, 100, "SwapTotal:      %8lu kB\n", memswlimit);
-			printme = lbuf;
-		} else if (startswith(line, "SwapFree:") && memswlimit > 0 && memswusage > 0) {
-			unsigned long swaptotal = memswlimit,
+		} else if (startswith(line, "SwapTotal:")) {
+                        if (memswlimit > 0) {
```
hallyn: Please collapse these cases nicer so that it's clear to the reader what's going on.  Since the second case will overwrite the first and memswlimit and hostswtotal are not used elsewhere this isn't wrong, but the reader has to go looking to make sure that those are in fact the case.

Actually, do you want the SwapTotal: line to show up in the swap_off == true case?

In either case, perhaps just

```
if (opts->swap_off == true)
   memswlimit = 0;
if (memswlimit > 0) { ... }

Comment (dmaj) [reply bindings.c]: I'm trying to rewrite the unsightly code right now. I wasn't happy with the implementation either.
To do that, I needed to understand the existing code even better. 
I came across a question. Is the order of the entries in "meminfo" fixed?
The reason for my question:
The variable "memswlimit" is set before the loop that processes the entries in "meminfo" starts.
In the line "SwapTotal" the value of "memswlimit" may be changed.
 if (hostswtotal < memswlimit)
              memswlimit = hostswtotal;
In the "SwapFree" line, the value of "memswlimit" is used in a formula.

Thus the result can change depending on the order of the entries "SwapTotal" and "SwapFree".

Comment (hallyn) [reply bindings.c]: If you look at fs/proc/meminfo.c in the kernel, you can see

```
        show_val_kb(m, "SwapTotal:      ", i.totalswap);
        show_val_kb(m, "SwapFree:       ", i.freeswap);
```
so the order is 'fixed', but there is no guarantee about the order (that is, a future code change can change it), so your concern is valid.

**Comment by hallyn** [comment]
Looks good to me, thanks .  I see you're still making changes, please comment when you're done and I'll merge.

**Comment by dmaj** [comment]
I've been through the code again. Now I don't notice anything anymore.  I adjusted some small things so that the formatting of the code is uniform.

Please take a look at the code again to be sure. I wrote my last C program about 20 years ago :-).

I won't change the code anymore. I have made all the changes I had on the list.

File: lxcfs.c [hunk 2]
```diff
@@ -792,9 +792,10 @@ static void usage()
 {
 	fprintf(stderr, "Usage:\n");
 	fprintf(stderr, "\n");
-	fprintf(stderr, "lxcfs [-f|-d] -l [-p pidfile] mountpoint\n");
+	fprintf(stderr, "lxcfs [-f|-d] -u -l -n [-p pidfile] mountpoint\n");
 	fprintf(stderr, "  -f running foreground by default; -d enable debug output \n");
 	fprintf(stderr, "  -l use loadavg \n");
+        fprintf(stderr, "  -u no swap \n");
```
hallyn: Please replace spaces with tab here.

lxcfs.c
```diff
@@ -929,7 +941,7 @@ int main(int argc, char *argv[])
 	}
 	if (swallow_option(&argc, argv, "-p", &v))
 		pidfile = v;
-
+
```
hallyn: This line is adding some spaces.

hallyn:
Apart from the two whitespace issues, looks great, thanks.

lxcfs.c
dmaj: Cleaned it up.

hallyn:
Argh, thanks - I'm sorry, I had missed one - the 'bool swap_off' also has leading spaces instead of tab.

Also the first line you add to README.md has just 4 spaces.

I will go ahead and fix those up, squash them all into one commit, and push.

Thank you.

dmaj:
Sorry for the many whitespaces and thanks for the patience.
I will configure netbeans correctly next time.
It was fun and I am happy that I could give something back to the community.
Thank you very much.

Issue #273, opened 2019-02-20
There is no cpuacct.usage_all file in CentOS 7.2，should use cpuacct.usage_percpu ?


lijiaocn:
OS:  CentOS Linux release 7.2.1511 (Core)
Kernel:  3.10.0-693.11.6.el7.x86_64
Commit:   `d8addb023a0e96b7d42e253703ba909bbc81dd26`

I can't find file cpuacct.usage_all in cgroup cpuacct controller directory, but find file cpuacct.usage_percpu.  cpuacct.usage_all has been instead by cpuacct.usage_percpu?

Now, lxcfs reads cgroup info from cpuacct.usage_all to generate file /proc/stat, if this file doesn't exist, it will just use host's /proc/stat file. So, we'll see host's cpu usage in container.

Open debug mode, will see the error: 

`bindings.c: 4821: proc_stat_read: proc_stat_read failed to read from cpuacct, falling back to the host's /proc/stat`

lijiaocn:
I try to solve this problem:  https://github.com/lxc/lxcfs/pull/274/files

hallyn:
You don't have cpuact.usage_all?  What kernel and distro are you on?

lijiaocn:
After upgrade  kernel to 4.20.12-1.el7.elrepo.x86_64, I find cpuact.usage_all, kernel 3.10 may be too old .

OS: CentOS Linux release 7.2.1511 (Core)
Kernel: 3.10.0-693.11.6.el7.x86_64

<img width="1254" alt="图片" src="https://user-images.githubusercontent.com/6396775/56499502-4237b880-6539-11e9-86a2-647d82d3bc11.png">

hallyn:
Ok then i'm tempted to close this issue and PR #275.  @brauner what do you think?

brauner:
Yeah, I think 3.13 at this point is our cut-off point.

PR #275, opened 2019-02-25
read cpu info from cpuacct.usage_percpu if no cpuacct.usage_all

lijiaocn:
Read cpu info from cpuacct.usage_percpu if no cpuacct.usage_all  #273  

Signed-off-by: lijiaocn <lijiaocn@foxmail.com>

### DIFF
--- a/bindings.c
+++ b/bindings.c
@@ -4071,12 +4071,16 @@ static inline int read_cpuacct_usage_from_percpu(
 
 	// usage_str format, 4 cores:
 	// 120190269376599 151483821809073 145465919648605 14450503259981
-	token = strtok(usage_str, delim);
-	while (NULL != token && j < cpucount)  {
-		sscanf(token,"%lu", &usage);
+	j = 0;
+	lxcfs_iterate_parts(token, usage_str, delim){
+		if (j >= cpucount) {
+			break;
+		};
+		if (sscanf(token,"%lu", &usage) != 1 ) {
+			goto err;
+		}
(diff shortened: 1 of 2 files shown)


bindings.c
```diff
@@ -4046,6 +4046,49 @@ static uint64_t get_reaper_age(pid_t pid)
...
+	}
+
+	// usage_str format, 4 cores:
+	// 120190269376599 151483821809073 145465919648605 14450503259981
+	token = strtok(usage_str, delim);
+	while (NULL != token && j < cpucount)  {
+		sscanf(token,"%lu", &usage);
+		cpu_usage[j].user = usage / 1000.0 / 1000 / 1000 * ticks_per_sec;
+		cpu_usage[j].system = 0;
+		token = strtok(NULL, delim);
```
brauner: We can't really use `strtok` since it's not thread-safe and `libfuse` runs threaded.

brauner: You can copy:
```c
#define lxc_iterate_parts(__iterator, __splitme, __separators)                  \
	for (char *__p = NULL, *__it = strtok_r(__splitme, __separators, &__p); \
	     (__iterator = __it);                                               \
	     __iterator = __it = strtok_r(NULL, __separators, &__p))
```
from https://github.com/lxc/lxc/blob/master/src/lxc/macro.h

lijiaocn: Thanks, I have fixed this problem with codes in lxc/macro.h,  at the same time add some codes to check `sscanf`'s return value.

brauner:
A version of this has been merged in another pull request, thank you!

PR #278, opened 2019-03-22
Support /sys/devices/system/cpu/online

wavezhang:
Support /sys/devices/system/cpu/online， to make applications that use sysconf (_SC_NPROCESSORS_ONLN) happy.

Partially resolve #88 .

PR #282, opened 2019-05-21
mount hook: mount /sys/devices/system/cpu/online

aither64:
Mount `/sys/devices/system/cpu/online` introduced by #278.


brauner:
Should this be opt-in? @stgraber?

stgraber:
Not sure why we'd make this opt-in when /proc/cpuinfo isn't, unless we know there's a downside of this being enabled for some reason.

brauner:
Nope, just wanted to check that you're fine with relying on this.

Issue #287, opened 2019-06-10
CPU views based on quotas: CPU usage is not correct


BurningXFlame:
Thank you for this excellent project and thank @aither64 for #260. This project really helps on a project I'm recently working on.
I met several issues when applying lxcfs, made some improvement and want to contribute to the community.
One of the issues is as follows.

CPU usage is not correct.

Test case:

- Start a container

``` shell
docker run -it -m 256m --cpu-period=100000 --cpu-quota=100000 \
    -v /var/lib/lxcfs/proc/cpuinfo:/proc/cpuinfo:ro \
    -v /var/lib/lxcfs/proc/meminfo:/proc/meminfo:ro \
    -v /var/lib/lxcfs/proc/stat:/proc/stat:ro \
    -v /var/lib/lxcfs/proc/swaps:/proc/swaps:ro \
    -v /var/lib/lxcfs/proc/uptime:/proc/uptime:ro \
    someimage /bin/bash
```
- Start a CPU-consuming process

``` shell
stress -c 1
```

- Expect CPU usage to be 100%

- But the CPU usage is 50% which is wrong

![issue_cpuUsage](https://user-images.githubusercontent.com/2758663/59171722-eff93800-8b76-11e9-9539-08e6525db62e.png)

BurningXFlame:
Proposed a resolution #290 


Effect: CPU usage is 100% as it should be.

![fixed_cpuUsage](https://user-images.githubusercontent.com/2758663/59179681-6c9b0f00-8b95-11e9-85cd-b5112f870a3c.png)

Issue #288, opened 2019-06-10
CPU views based on quotas: CPU usage is not correct in partial CPU cases when quota/period doesn't yield an integer


BurningXFlame:
Thank you for this excellent project and thank @aither64 for #260. This project really helps on a project I'm recently working on.
I met several issues when applying lxcfs, made some improvement and want to contribute to the community.
One of the issues is as follows.

CPU usage is not correct in partial CPU cases when quota/period don't yield an integer

Test case:

- Start a container

``` shell
docker run -it -m 256m --cpu-period=100000 --cpu-quota=150000 \
    -v /var/lib/lxcfs/proc/cpuinfo:/proc/cpuinfo:ro \
    -v /var/lib/lxcfs/proc/meminfo:/proc/meminfo:ro \
    -v /var/lib/lxcfs/proc/stat:/proc/stat:ro \
    -v /var/lib/lxcfs/proc/swaps:/proc/swaps:ro \
    -v /var/lib/lxcfs/proc/uptime:/proc/uptime:ro \
    someimage /bin/bash
```
- Start CPU-consuming processes

``` shell
stress -c 2
```

- Expect CPU usage to be 100%

- But the CPU usage is 75% which is wrong
  - total
![issue_cpuUsage_partialCpu](https://user-images.githubusercontent.com/2758663/59171935-02c03c80-8b78-11e9-8f3b-1ec597466355.png)
  - per core
![issue_cpuUsage_partialCpu_perCore](https://user-images.githubusercontent.com/2758663/59171957-179cd000-8b78-11e9-9d40-8e58fba2e07a.png)

BurningXFlame:
Proposed a resolution #290

Effect: CPU usage is 100% as it should be.

- total
![fixed_cpuUsage_partialCpu](https://user-images.githubusercontent.com/2758663/59179838-d87d7780-8b95-11e9-84aa-26c1ec643a20.png)
- per core
![fixed_cpuUsage_partialCpu_perCore](https://user-images.githubusercontent.com/2758663/59179868-eaf7b100-8b95-11e9-92d9-793cb8985d5e.png)

Issue #286, opened 2019-06-10
CPU views based on quotas: fall back to cpuacct.usage_percpu if cpuacct.usage_all not exists


BurningXFlame:
Thank you for this excellent project and thank @aither64 for #260. This project really helps on a project I'm recently working on.
I met several issues when applying lxcfs, made some improvement and want to contribute to the community. 
One of the issues is as follows.

#260 reads cpuacct.usage_all. But cpuacct.usage_all doesn't exist on our hosts for containers.
A resolution would be falling back to cpuacct.usage_percpu if cpuacct.usage_all not exists.

BurningXFlame:
Proposed a resolution #290

PR #290, opened 2019-06-10
CPU views based on quotas: several improvements

BurningXFlame:
Thank you for this excellent project and thank @aither64 for #260. This project really helps on a project I'm recently working on.
I met several issues when applying lxcfs, made some improvement and want to contribute to the community.

As follows are issues, and corresponding causes and resolutions.

- #286 CPU views based on quotas: fall back to cpuacct.usage_percpu if cpuacct.usage_all not exists
  - Resolution
if cpuacct.usage_all not exist, read cpuacct.usage_percpu instead, and  convert it to the format of cpuacct.usage_all
  - Note
I'm aware that there's anthor PR #275 which aims to achieve the same goal. But the resolution proposed here is elegant, more understandable and has less impact to the existing flow.

- #287 CPU usage is not correct
  - Cause
The statistic logic of CPU usage is not correct. When computing the CPU usage of a container, there's always one more core, which the container doesn't use, being included. 
Say, the correct CPU usage should be U/T where U stands for the capacity used by the container, and T stands for the total capacity assigned to the container.  But the calculated usage is U/(T+D) where D stands for the capacity of the extra one more core. So the calculated usage is lower than it should be.  
  - Resolution
As cause being explained, a resolution is straightforward.
Please refer to #287 for effect.

- #288 CPU usage is not correct in partial CPU cases when quota/period doesn't yield an integer
  - Cause
When quota/period doesn't yield an integer, the fractional part is treated as a whole core. But the quota remains unchanged. 
Say, the correct CPU usage should be U/T where U stands for the capacity used by the container, and T stands for the total capacity assigned to the container.  But the calculated usage is U/(T+D) where D stands for the extra capacity as a result of ceiling quota/period. So the calculated usage is lower than it should be.  
  - Resolution
Revise CPU usage to match the exact quota.
Please refer to #288 for effect.


brauner:
Hey, thanks for the patch. Can you please sign-off your commits?

bindings.c
```diff
@@ -1155,6 +1156,7 @@ bool cgfs_get_value(const char *controller, const char *cgroup, const char *file
 	char *fnam, *tmpc;
 
 	tmpc = find_mounted_controller(controller, &cfd);
+	lxcfs_debug("tmpc: %s\n", tmpc);
 	if (!tmpc)
 		return false;
```
brauner: He, this would be a NULL-dereference if `find_mounted_controller()` fails, no? :)

bindings.c
```diff
@@ -1172,6 +1175,7 @@ bool cgfs_get_value(const char *controller, const char *cgroup, const char *file
...
 
 	*value = slurp_file(fnam, fd);
+	lxcfs_debug("*value: %s\n", *value);
```
brauner: NULL dereference on `slurp_file()` error.

bindings.c
```diff
@@ -3679,6 +3683,35 @@ int max_cpu_count(const char *cg)
 	return rv;
 }
 
+/*
+ * Return the exact number of visible CPUs based on CPU quotas.
+ * If there is no quota set, zero is returned.
+ */
+double exact_cpu_count(const char *cg)
```
brauner: should be `static`

bindings.c
```diff
@@ -4034,20 +4067,47 @@ static int read_cpuacct_usage_all(char *cg, char *cpuset, struct cpuacct_usage *
...
-		lxcfs_debug(
+		lxcfs_v(
 			"%s\n",
 			"read_cpuacct_usage_all failed to determine number of clock ticks "
 			"in a second");
 		return -1;
 	}
 
 	cpu_usage = malloc(sizeof(struct cpuacct_usage) * cpucount);
+	bzero(cpu_usage,sizeof(struct cpuacct_usage) * cpucount);
```
brauner: `bzero` is deprecated. Please replace with `memset(cpu_usage, 0, sizeof(...))`

bindings.c
```diff
@@ -4034,20 +4067,47 @@ static int read_cpuacct_usage_all(char *cg, char *cpuset, struct cpuacct_usage *
...
+		// read cpuacct.usage_percpu instead
+		lxcfs_v("reading cpuacct.usage_percpu instead%s\n", "");
+		if (!cgfs_get_value("cpuacct", cg, "cpuacct.usage_percpu", &usage_str)) {
+			rv = -1;
+			goto err;
+		}
+
+		// convert cpuacct.usage_percpu into cpuacct.usage_all
+		lxcfs_v("converting cpuacct.usage_percpu into cpuacct.usage_all%s\n", "");
+		StringBuilder *sb = sb_create();
```
brauner: We don't do camelcase, not even for new types. So this should either by `stringbuilder` or `string_builder`

bindings.c
```diff
@@ -4034,20 +4067,47 @@ static int read_cpuacct_usage_all(char *cg, char *cpuset, struct cpuacct_usage *
...
+		sb_append(sb, "cpu user system\n");
+		int i = 0;
+		while(sscanf(usage_str + read_pos, "%lu %n", &cg_user, &read_cnt) > 0){
```
brauner: Missing space between `while` keyword and opening `(`as well as missing space after `)` and `{`

bindings.c
```diff
@@ -4623,7 +4701,7 @@ static int cpuview_proc_stat(const char *cg, const char *cpuset, struct cpuacct_
 			stat_node->view[curcpu].user = stat_node->usage[curcpu].user;
 			stat_node->view[curcpu].system = stat_node->usage[curcpu].system;
 			stat_node->view[curcpu].idle = stat_node->usage[curcpu].idle;
-
+
```
brauner: Hm?

BurningXFlame:
@brauner Thank you. I've made requested changes.

sb.c
```diff
@@ -103,17 +103,17 @@ int sb_appendf(StringBuilder *sb, const char *format, ...)
...
  */
-char *sb_concat(StringBuilder *sb)
+char *sb_concat(stringbuilder *sb)
 {
 	char			*buf = NULL;
 	char			*c = NULL;
-	StringFragment	*frag = NULL;
+	stringfragment	*frag = NULL;
 
 	buf = (char *) malloc((sb->length + 1) * sizeof(char));
```
hallyn: In read_cpuacct_usage_all() you can call sb_concat(NULL).  Here you deference sb->length without checking.

hallyn: Please squash these commits, it will make this easier to analyze.

BurningXFlame:
Thanks. I've added validation to stringbuilder and where it's used. Also squashed commits.

hallyn:
Thanks, +1 from me.  @brauner I'll leave it to you to merge when you're happy :)

sb.c
```diff
@@ -0,0 +1,168 @@
...
+		return SB_FAILURE;
+
+	int				length = 0;
+	stringfragment	*frag = NULL;
+
+	if (NULL == str || '\0' == *str)
+		return sb->length;
+
+	length = strlen(str);
+	frag = (stringfragment*) malloc(sizeof(stringfragment) + (sizeof(char) * length));
```
brauner: Can you please remove all casts from allocation functions. Something we don't do at all. :)

BurningXFlame:
Thanks. I've made requested changes.

bindings.c
```diff
@@ -4045,9 +4075,48 @@ static int read_cpuacct_usage_all(char *cg, char *cpuset, struct cpuacct_usage *
...
+			}
+			i++;
+			read_pos += read_cnt;
+		}
+
+		free(usage_str);
+		usage_str = sb_concat(sb);
+		sb_free(sb);
+		read_pos=0;
+		read_cnt=0;
```
Blub: The next access to `read_pos` includes a read (`+=`), which makes me really uncomfortable, despite the only thing happening before being an initialization to `0`, because that can now easily get lost in patch contexts.
I'd rather have `read_pos/cnt` shadowed in this block, and the initialization of the "outer" `read_pos` moved down.

sb.h
```diff
@@ -0,0 +1,56 @@
...
+#ifndef SB_H
+#define SB_H
+
+#define SB_FAILURE				-1
+#define SB_MAX_FRAG_LENGTH		4096
+
+typedef struct _stringfragment {
+	struct _stringfragment	*next;
+	int						length;
+	char					*str;
```
Blub: This is confusing as `str` is not used as a pointer. Please make this either just `char str` or `str[0]` or `str[]` (the latter two will each make different compilers unhappy...
@brauner, @hallyn ^ do we already have a style precedent for this?

sb.h
```diff
@@ -0,0 +1,56 @@
...
+
+#ifndef SB_H
+#define SB_H
+
+#define SB_FAILURE				-1
+#define SB_MAX_FRAG_LENGTH		4096
+
+typedef struct _stringfragment {
+	struct _stringfragment	*next;
+	int						length;
```
Blub: I'd like to see all length types to be `size_t` (except in temporary local variables where they come from a function returning a signed type).
I just noticed some signedness issues in the existing code base which we should address after this PR (it would otherwise merge-conflict) (see `clang` with `-Wtautological-compare`)

BurningXFlame:
Thanks. I've made requested changes.

It turns out that the stringbuilder I'm using is not ready. I've revised it several times. It would be better to use a mature one, for example string_utils you mentioned.

BurningXFlame:
@Blub Regarding stringbuilder, what do you think?
How about -
- Merge this PR
- Merge string_utils you mentioned
- Me fire another PR to use string_utils instead of stringbuilder

Blub:
Actually there's no direct `append_string()` function in string_utils yet, however, in lxcfs there's already a `must_strcat_pid()` function of which the `must_strcat()` part could be factored out. If I'm not mistaken that function does exactly what you need: it formats an integer into a temporary buffer (which sb_appendf does as well), then appends it to an existing string, reallocating it as needed.

h0tw1r3:
Been running this on Proxmox in a pre-production environment for the last week. :+1: 
Plan on rolling it out to all environments soon.

brauner:
> Actually there's no direct `append_string()` function in string_utils yet, however, in lxcfs there's already a `must_strcat_pid()` function of which the `must_strcat()` part could be factored out. If I'm not mistaken that function does exactly what you need: it formats an integer into a temporary buffer (which sb_appendf does as well), then appends it to an existing string, reallocating it as needed.
> 
> (PS: sorry for the delay...)

Agreed. And with that change done I think we're ready to merge.

BurningXFlame:
Abstracted must_strcat from must_strcat_pid, and replaced stringbuild with must_strcat.

brauner:
@Blub are you happy with this change?

bindings.c
```diff
@@ -4045,11 +4090,41 @@ static int read_cpuacct_usage_all(char *cg, char *cpuset, struct cpuacct_usage *
 	if (!cpu_usage)
 		return -ENOMEM;
 
+	memset(cpu_usage, 0, sizeof(struct cpuacct_usage) * cpucount);
 	if (!cgfs_get_value("cpuacct", cg, "cpuacct.usage_all", &usage_str)) {
-		rv = -1;
-		goto err;
+		lxcfs_v("failed to read cpuacct.usage_all%s\n", "");
```
Blub: What's the reason behind the empty string added there? This should compile fine without?

bindings.c
```diff
@@ -4045,11 +4090,41 @@ static int read_cpuacct_usage_all(char *cg, char *cpuset, struct cpuacct_usage *
...
+
+		// read cpuacct.usage_percpu instead
+		lxcfs_v("reading cpuacct.usage_percpu instead%s\n", "");
```
Blub: Maybe combine the two messages into one?

bindings.c
```diff
@@ -4045,11 +4090,41 @@ static int read_cpuacct_usage_all(char *cg, char *cpuset, struct cpuacct_usage *
...
+		}
+		lxcfs_v("usage_str: %s\n", usage_str);
+
+		// convert cpuacct.usage_percpu into cpuacct.usage_all
+		lxcfs_v("converting cpuacct.usage_percpu into cpuacct.usage_all%s\n", "");
+		
+		char *data = NULL;
+		size_t sz = 0, asz = 0;
+
+		must_strcat(&data, &sz, &asz, "cpu user system\n", "");
```
Blub: The last parameter should not be necessary.

bindings.c
```diff
@@ -4563,14 +4638,14 @@ static int cpuview_proc_stat(const char *cg, const char *cpuset, struct cpuacct_
 		threshold = total_sum / cpu_cnt * max_cpus;
 
 		for (curcpu = 0, i = -1; curcpu < nprocs; curcpu++) {
-			if (i == max_cpus)
```
Blub: Do you have a reason for moving this block? I don't see `i` used for anything other than counting?

BurningXFlame: To fix issue #287 CPU usage is not correct.  The cause and resolution are explained above.

bindings.c
BurningXFlame: Otherwise, the compiler complains:

```
macro.h:7:25: error: expected expression before ')' token
    __func__, __VA_ARGS__);                                \
                         ^
macro.h:10:34: note: in expansion of macro 'lxcfs_debug_stream'
 #define lxcfs_error(format, ...) lxcfs_debug_stream(stderr, format, __VA_ARGS__)
                                  ^
macro.h:19:30: note: in expansion of macro 'lxcfs_error'
 #define lxcfs_v(format, ...) lxcfs_error(format, __VA_ARGS__);
                              ^
bindings.c:4104:3: note: in expansion of macro 'lxcfs_v'
   lxcfs_v("converting cpuacct.usage_percpu into cpuacct.usage_all\n");
   ^
```

BurningXFlame:
@Blub commented respectively

brauner:
@Blub, once you're happy, please merge.

brauner:
> @Blub commented respectively.
> 
> Besides, I'd like to keep the 2 commits, so that I've a proven-to-work version to roll back if something went wrong.

We don't do things like that unfortunately. But you can keep a local copy of that branch around and I can too if you want me to. I'm going to squash those commits into a single one.

brauner:
Thanks for the work and the patience. :)

PR #293, opened 2019-07-18
bindings: enable higher precision uptime

### LINKED ISSUE #292: /proc/uptime unnecessarily truncates microseconds
lxfs: 3.0.3-0ubuntu1

When looking at a /proc/uptime in a container microsecond granularity is automatically truncated to seconds. This makes it hard to properly calculdate start times of processes in microseconds relative to container start time. 

The code related to this truncation is:

bindings.c:proc_uptime_read:5061 which only attempts to represent uptime in seconds @
        total_len = snprintf(d->buf, d->buflen, "%"PRIu64".00 %"PRIu64".00\n", reaperage, idletime);

There are additional comments that describe the intent behind only using seconds as the granularity of get_reaper_age:

./bindings.c:get_reaper_age:~3991 
                /* We could make this more precise here by using the tv_nsec
                 * field in the timespec struct and convert it to milliseconds
                 * and then create a double for the seconds and milliseconds but
                 * that seems more work than it is worth.
                 */

```
# example from within the container
root@x1:~# cat /proc/uptime 
91622.00 91520.00

# on the host we do get more specific uptime
1031969.86 8152657.65
```

brauner:
I don't know for what reason anyone but care but I guess ok.

Closes #292.

### DIFF
--- a/bindings.c
+++ b/bindings.c
@@ -4010,10 +4010,11 @@ static uint64_t get_reaper_start_time(pid_t pid)
 	return starttime;
 }
 
-static uint64_t get_reaper_start_time_in_sec(pid_t pid)
+static double get_reaper_start_time_in_sec(pid_t pid)
 {
-	uint64_t clockticks;
-	int64_t ticks_per_sec;
+	uint64_t clockticks, ticks_per_sec;
+	int64_t ret;
+	double res = 0;
 
 	clockticks = get_reaper_start_time(pid);
 	if (clockticks == 0 && errno == EINVAL) {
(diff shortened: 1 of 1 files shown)


bindings.c
```diff
@@ -4029,12 +4030,15 @@ static uint64_t get_reaper_start_time_in_sec(pid_t pid)
...
+	return res;
 }
 
-static uint64_t get_reaper_age(pid_t pid)
+static double get_reaper_age(pid_t pid)
 {
-	uint64_t procstart, uptime, procage;
+	uint64_t uptime_ms;
+	double procstart, procage;
+	// double procstart, uptime, procage;
```
hallyn: You'd never let anyone get away with that :-)

brauner: Oh yeah, that's so sloppy. I'm sorry. :) Will fix!

brauner: Should be fixed. :)

PR #304, opened 2019-08-29
Fix proc_diskstats_read function

time-river:
`# cat /proc/diskstats` sometime is none in container although using lxcfs. The reason is that incorrect way of getting printed information.

We could get block device information include device name, physical sector size, scheduler name from sysfs (`/sys/dev/block/<major:minor>/{uevent, queue/hw_sector_size, queue/scheduler}` by `major:minor`. No need to traverse `/proc/diskstats` to decide which disk information should be printed.

There are many scheduler in current linux kernel, for example: cfq, bfq, none, mq-deadline. cgroup block controller file `blkio.[a-z_]+` is only for cfq, and bfq is `blkio.bfq.[a-z_]+`. More general, throttle works at generic block layer suitable for any scheduler. So We could get `io_service` and `io_service_bytes` from throttle instead of specific scheduler.

Further, `blkio.*_recursive` show the same information as their non-recursive counterparts but include stats from all the descendant cgroups according to [linux doc](https://www.kernel.org/doc/Documentation/cgroup-v1/blkio-controller.txt), and there is cgroup namespace. We collect data from 'blkio.*_recursive' for more accuracy.

### DIFF
--- a/bindings.c
+++ b/bindings.c
@@ -3322,6 +3322,62 @@ static void parse_memstat(char *memstat, unsigned long *cached,
 	}
 }
 
+static bool get_blk_sched_name(unsigned int major, unsigned int minor,
+				char *scheduler, size_t n)
+{
+	char path[PATH_MAX] = {0};
+	char *line = NULL, *alpha;
+	size_t len = 0, i = 0;
+	FILE *file = NULL;
+	bool retval = false;
+
+	len = snprintf(path, sizeof(path),
+			"/sys/dev/block/%u:%u/queue/scheduler", major, minor);
+	if (len < 0) {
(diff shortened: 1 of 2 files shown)


brauner:
If pulled the patch series into my local repo and am looking through it. In general I think the sysfs parts are worth it. But the series removes looping through all the devices listed in `/proc/diskstats` and only looks at a single `major`, `minor` pair. That seems wrong to me. If I'd be convinced it's correct right away, I would've ported it to the new, refactored lxcfs version myself. But since I'm not entirely sure this is right I'd ask you to port this onto master and I'll review it. Thanks and sorry for the hazzle.

PR #305, opened 2019-09-06
fix /sys/devices/system/cpu/online in cpuset not correct bug

yinhongbo:
When the container runs in cpuset and cpu quota, for example, cpu quota is limited to 8 cores, cpuset is limited to 1-3, 4-5, then there will be problems. The core number obtained by /proc/cpuinfo is 4 cores, and /sys /devices/system/cpu/online gets 8 cores.
The reason is that /sys/devices/system/cpu/online does not take into account the situation of cpuset. So submit this patch.

Test case:
`docker run -it -d  --cpuset-cpus="1-2,3-4" \
--cpu-period=100000 \
--cpu-quota=800000 \
--memory 2048m \
-v /root/tce/lxcfs/run/lxcfs/sys/devices/system/cpu/online:/sys/devices/system/cpu/online \
-v /root/tce/lxcfs/run/lxcfs/proc/cpuinfo:/proc/cpuinfo:rw \
debian:latest bash
`

### DIFF
--- a/bindings.h
+++ b/bindings.h
@@ -82,5 +82,6 @@ extern char *get_cpuset(const char *cg);
 extern bool use_cpuview(const char *cg);
 extern int max_cpu_count(const char *cg);
 extern void do_release_file_info(struct fuse_file_info *fi);
+extern int cpu_number_in_cpuset(const char *cpuset);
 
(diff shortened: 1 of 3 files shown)


brauner:
Thanks.

Issue #301, opened 2019-08-21
/sys/devices/system/cpu/online not container aware 


ilhaan:
I have noticed that containers see all cores when I inspect `/sys/devices/system/cpu/online`, even though they have been restricted to a single core. `/proc/cpuinfo` seems to be reporting correctly. 

For example, I have a container `test`: 
```
root@server:~# lxc exec test -- grep -c "processor" /proc/cpuinfo 
1
```
Which is the expected output. However. 
```
root@server:~# lxc exec test -- cat /sys/devices/system/cpu/online
0-55
```
I tested this using the following snap installed versions of LXD: 
* 3.0.4 (from the `3.0/stable` channel)
* 3.16 (from the `3.16/stable` channel)

More info from my server: 
```
root@server:~# lsb_release -a 
No LSB modules are available.
Distributor ID:	Ubuntu
Description:	Ubuntu 18.04.3 LTS
Release:	18.04
Codename:	bionic
root@server:~# uname -a 
Linux v45 4.15.0-58-generic #64-Ubuntu SMP Tue Aug 6 11:12:41 UTC 2019 x86_64 x86_64 x86_64 GNU/Linux
root@server:~# snap --version 
snap    2.40
snapd   2.40
series  16
ubuntu  18.04
kernel  4.15.0-58-generic
```
Seems like there was a [merge](https://github.com/lxc/lxcfs/pull/282) in May 2019 to address this. The comments on this merge make it seem like this should be enabled by default. 

I discovered this while setting up a Kubernetes cluster with [Rancher](https://rancher.com/) where I used LXD containers as nodes and [posted an issue in the rancher repo](https://github.com/rancher/rancher/issues/21057). Rancher seems to be using `/sys/` to enumerate CPU and memory resources on nodes, which is why I need LXCFS to fix my problem. 

I also [posted about this on the LXD forum](https://discuss.linuxcontainers.org/t/lxcfs-and-rancher-resource-enumeration/5549). 

Please let me know if  you need me to provide additional information.

stgraber:
Yeah, there is a bug in the generation of the `online` value...

```
root@shell01:~# grep lxcfs /proc/mounts
lxcfs /proc/cpuinfo fuse.lxcfs rw,nosuid,nodev,relatime,user_id=0,group_id=0,allow_other 0 0
lxcfs /proc/diskstats fuse.lxcfs rw,nosuid,nodev,relatime,user_id=0,group_id=0,allow_other 0 0
[...]

PR #307, opened 2019-09-27
fix the problem of counting the number of CPUs when using cpuset and …

yinhongbo:
This PR mainly fixes the problem of counting the number of CPUs when using cpuset and cpuset or using cpuset alone.
When using only cpuset

```
Docker run -it --rm \
--cpuset-cpus="0-4,6" \
--memory 2048m \
-v /var/run/lxcfs/sys/devices/system/cpu/online:/sys/devices/system/cpu/online \
-v /var/run/lxcfs/proc/cpuinfo:/proc/cpuinfo:rw \
Ningx bash
```
Or when Cpuset and cpu quota are used at the same time, but the number of CPUs that cpuset can use is less than cpu quota.
```
Docker run -it --rm \
--cpu-period=100000 \
--cpu-quota=800000 \
--cpuset-cpus="0-4,6" \
--memory 2048m \
-v /var/run/lxcfs/sys/devices/system/cpu/online:/sys/devices/system/cpu/online \
-v /var/run/lxcfs/proc/cpuinfo:/proc/cpuinfo:rw \
Ningx bash
```
Getting the number of CPUs in the container will be unsatisfactory.

Therefore, cpuset is introduced as the basis for the quantity judgment when calculating the maximum available CPU number.

This PR can also solve these ISSUE（such as [331](https://github.com/lxc/lxcfs/issues/301)） problems.

### DIFF
--- a/bindings.c
+++ b/bindings.c
@@ -3669,6 +3669,8 @@ int max_cpu_count(const char *cg)
 {
 	int rv, nprocs;
 	int64_t cfs_quota, cfs_period;
+	int nr_cpus_in_cpuset = 0;
+	char *cpuset = NULL;
 
 	if (!read_cpu_cfs_param(cg, "quota", &cfs_quota))
 		return 0;
@@ -3676,8 +3678,16 @@ int max_cpu_count(const char *cg)
 	if (!read_cpu_cfs_param(cg, "period", &cfs_period))
 		return 0;
 
-	if (cfs_quota <= 0 || cfs_period <= 0)
+	cpuset = get_cpuset(cg);
+	if (cpuset)
+		nr_cpus_in_cpuset = cpu_number_in_cpuset(cpuset);
+
+	if (cfs_quota <= 0 || cfs_period <= 0){
+		if (nr_cpus_in_cpuset > 0)
+			return nr_cpus_in_cpuset;
+
 		return 0;
+	}
 
 	rv = cfs_quota / cfs_period;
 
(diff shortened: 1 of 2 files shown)


bindings.c
```diff
@@ -3669,15 +3669,25 @@ int max_cpu_count(const char *cg)
...
 	if (!read_cpu_cfs_param(cg, "period", &cfs_period))
 		return 0;
 
-	if (cfs_quota <= 0 || cfs_period <= 0)
+	cpuset = get_cpuset(cg);
+	if (cpuset)
+		nr_cpus_in_cpuset = cpu_number_in_cpuset(cpuset);
+
+	if (cfs_quota <= 0 || cfs_period <= 0){
+		if(nr_cpus_in_cpuset > 0)
```
brauner: missing space between `if` and `(`.

bindings.c
```diff
@@ -3692,6 +3702,10 @@ int max_cpu_count(const char *cg)
 	if (rv > nprocs)
 		rv = nprocs;
 
+	/* use min value in cpu quota and cpuset */
+	if(nr_cpus_in_cpuset > 0 && nr_cpus_in_cpuset < rv)
```
brauner: missing space between `if` and `(`.

bindings.c
yinhongbo: addressed

bindings.c
yinhongbo: addressed

yinhongbo:
ping @brauner

hallyn:
Thanks - lgtm, but it would be nice to have the two commits squashed.  Please let me know if you'd prefer I do it.

yinhongbo:
> Thanks - lgtm, but it would be nice to have the two commits squashed. Please let me know if you'd prefer I do it.

Sorry, I didn’t pay attention to this issue during the holiday. I will merge two submissions when I encounter the next PR.

PR #308, opened 2019-10-09
fixed a problem with server stability degradation caused by cgroup dy…

yinhongbo:
When we troubleshoot the problem of server network jitter at one time, the reason for the positioning is that the kernel is traversing the cgroup, and the main time is small in the memcg_stat_show function. Our kernel engineer obtains lxcfs by grabbing the stack and triggers this problem. After combining lxcfs code analysis and positioning, it is determined that the code disappears after reload lxcfs.

kernel stack：
```
    COMMAND: lxcfs PID: 3223863 LATENCY: 55ms
     trace_irqoff_record+0x12b/0x1b0 [trace_irqoff]
     trace_irqoff_hrtimer_handler+0x97/0x99 [trace_irqoff]
     __hrtimer_run_queues+0xdc/0x220
     hrtimer_interrupt+0xa6/0x1f0
     smp_apic_timer_interrupt+0x62/0x120
     apic_timer_interrupt+0x7d/0x90
     memcg_stat_show+0x27a/0x460
     seq_read+0x11f/0x3f0
     __vfs_read+0x33/0x160
     vfs_read+0x91/0x130
     SyS_read+0x52/0xc0
     do_syscall_64+0x68/0x100
     entry_SYSCALL_64_after_hwframe+0x3d/0xa2
```

this pr fixed a problem with server stability degradation caused by cgroup dying traversal. This is a very bad problem.
when initpid == 1, getting meminfo will traverses all the memory cgroup.
a certain probability causes server stability to drop, such as network jitter.

cgroup dying examples are as follows。userspace has 1810 memcgroup, but kernel has 110588 memory cgroup

```
root@XX-YY-ZZ-AA:~# lscgroup | grep memory | wc -l
1810
root@XX-YY-ZZ-AA:~# grep memory /proc/cgroups
memory	7	110588	1
````

more questions about cgroup dying.
Can refer to https://lwn.net/Articles/787614/?spm=a2c4e.10696291.0.0.105919a4uX5P3F

### DIFF
--- a/bindings.c
+++ b/bindings.c
@@ -738,7 +738,7 @@ static bool is_shared_pidns(pid_t pid)
 	if (pid != 1)
 		return false;
 
-	if (in_same_ns(pid, getpid(), "pid") == -EINVAL)
+	if (in_same_namespace(pid, getpid(), "pid") == -EINVAL)
 		return true;
 
 	return false;
@@ -2045,7 +2045,7 @@ int cg_getattr(const char *path, struct stat *sb)
 	}
 
 	pid_t initpid = lookup_initpid_in_store(fc->pid);
-	if (initpid <= 0 || is_shared_pidns(initpid))
+	if (initpid <= 1 || is_shared_pidns(initpid))
 		initpid = fc->pid;
 	/* check that cgcopy is either a child cgroup of cgdir, or listed in its keys.
 	 * Then check that caller's cgroup is under path if last is a child
@@ -2130,7 +2130,7 @@ int cg_opendir(const char *path, struct fuse_file_info *fi)
 	}
 
 	pid_t initpid = lookup_initpid_in_store(fc->pid);
-	if (initpid <= 0 || is_shared_pidns(initpid))
+	if (initpid <= 1 || is_shared_pidns(initpid))
 		initpid = fc->pid;
 	if (cgroup) {
 		if (!caller_may_see_dir(initpid, controller, cgroup))
@@ -2190,7 +2190,7 @@ int cg_readdir(const char *path, void *buf, fuse_fill_dir_t filler, off_t offset
 	}
 
 	pid_t initpid = lookup_initpid_in_store(fc->pid);
-	if (initpid <= 0 || is_shared_pidns(initpid))
+	if (initpid <= 1 || is_shared_pidns(initpid))
 		initpid = fc->pid;
 	if (!caller_is_in_ancestor(initpid, d->controller, d->cgroup, &nextcg)) {
 		if (nextcg) {
@@ -2301,7 +2301,7 @@ int cg_open(const char *path, struct fuse_file_info *fi)
 	free_key(k);
 
 	pid_t initpid = lookup_initpid_in_store(fc->pid);
-	if (initpid <= 0 || is_shared_pidns(initpid))
+	if (initpid <= 1 || is_shared_pidns(initpid))
 		initpid = fc->pid;
 	if (!caller_may_see_dir(initpid, controller, path1)) {
 		ret = -ENOENT;
@@ -2379,7 +2379,7 @@ int cg_access(const char *path, int mode)
 	free_key(k);
 
 	pid_t initpid = lookup_initpid_in_store(fc->pid);
-	if (initpid <= 0 || is_shared_pidns(initpid))
+	if (initpid <= 1 || is_shared_pidns(initpid))
 		initpid = fc->pid;
 	if (!caller_may_see_dir(initpid, controller, path1)) {
 		ret = -ENOENT;
@@ -3270,7 +3270,7 @@ int cg_mkdir(const char *path, mode_t mode)
 		path1 = cgdir;
 
 	pid_t initpid = lookup_initpid_in_store(fc->pid);
-	if (initpid <= 0 || is_shared_pidns(initpid))
+	if (initpid <= 1 || is_shared_pidns(initpid))
 		initpid = fc->pid;
 	if (!caller_is_in_ancestor(initpid, controller, path1, &next)) {
 		if (!next)
@@ -3328,7 +3328,7 @@ int cg_rmdir(const char *path)
 	}
 
 	pid_t initpid = lookup_initpid_in_store(fc->pid);
-	if (initpid <= 0 || is_shared_pidns(initpid))
+	if (initpid <= 1 || is_shared_pidns(initpid))
 		initpid = fc->pid;
 	if (!caller_is_in_ancestor(initpid, controller, cgroup, &next)) {
 		if (!last || (next && (strcmp(next, last) == 0)))
@@ -3532,7 +3532,7 @@ static int proc_meminfo_read(char *buf, size_t size, off_t offset,
 	}
 
 	pid_t initpid = lookup_initpid_in_store(fc->pid);
-	if (initpid <= 0 || is_shared_pidns(initpid))
+	if (initpid <= 1 || is_shared_pidns(initpid))
 		initpid = fc->pid;
 	cg = get_pid_cgroup(initpid, "memory");
 	if (!cg)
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

**Comment by brauner** [comment]
@Blub, @hallyn, @stgraber, thoughts?

**Comment by stgraber** [comment]
Hmm, so in this case initpid is the pid in the initial pid namespace right?
So the problem is when the real pid1 accesses lxcfs?
Why would that ever happen?

It certainly seems fine to return the real data for host pid 1 as that process is typically not in any cgroup itself, but it still feels weird that it would ever access lxcfs.

**Comment by yinhongbo** [comment]
> Hmm, so in this case initpid is the pid in the initial pid namespace right?
> So the problem is when the real pid1 accesses lxcfs?
> Why would that ever happen?
> 
> It certainly seems fine to return the real data for host pid 1 as that process is typically not in any cgroup itself, but it still feels weird that it would ever access lxcfs.

@stgraber When using docker, in some scenarios, the --pid = host mode is used, in which all processes of the managed machine are visible to the container.
Especially in K8S+docker, some daemonsets used to manage the host will use this mode. At this time, the initpid of the container is 1.

**Comment by stgraber** [comment]
Hmm, but in that case, your fix would actually lead to showing the wrong data to the container wouldn't it?

Because a container with `--pid=host` could and likely will still have cgroup limits applied to itself and so lxcfs is still needed to show the proper values in proc files.

**Comment by yinhongbo** [comment]
> Hmm, but in that case, your fix would actually lead to showing the wrong data to the container wouldn't it?
> 
> Because a container with `--pid=host` could and likely will still have cgroup limits applied to itself and so lxcfs is still needed to show the proper values in proc files.

@stgraber Yes, when --pid=host is enabled and the memory used by the container is limited, there will be a problem of returning the wrong cgroup memory data.
I am also thinking about this problem, lxcfs now use the way to find the initpid of the container may have some drawbacks, I am also trying other ways to correct this drawback.
At present, I suggest that this method can be used to avoid the problem of traversing all cgroups due to the calculation of memory usage in initpid=1. Because of the problem of traversal cgroup drying, this problem will cause server jitter, especially the network performance.
When I find a more suitable solution, I will continue to submit new PRs.

**Comment by yinhongbo** [comment]
> Hmm, but in that case, your fix would actually lead to showing the wrong data to the container wouldn't it?
> 
> Because a container with `--pid=host` could and likely will still have cgroup limits applied to itself and so lxcfs is still needed to show the proper values in proc files.

@stgraber as mentioned above, I found that when the container is running in `--pid=host` mode, initpid=1. At this time, the container can see all the processes of the host. We should correct the initpid to access the pid of the fuse (fc->pid). Through this process, We can get the corresponding cgroup path correctly.
diff code show as below:
```
-       if (initpid <= 0)
+       if (initpid <= 0 || initpid == 1)
                initpid = fc->pid;
        cg = get_pid_cgroup(initpid, "memory");
```
test script
```
docker run -it --rm \
--pid=host \
--memory 2048m \
-v /run/lxcfs/proc/meminfo:/proc/meminfo:rw debian:latest bash
```
The test results are as follows:
before fixing:
```

0200dfb77e0e(@:):/# free -h
             total       used       free     shared    buffers     cached
Mem:           62G       5.6G        56G       621M       340M       2.8G
-/+ buffers/cache:       2.5G        59G
Swap:           0B         0B         0B
```
After modification
```
7b89f3fada7e(@:):/# free -h
             total       used       free     shared    buffers     cached
Mem:          2.0G       1.4M       2.0G         0B         0B         0B
-/+ buffers/cache:       1.4M       2.0G
Swap:           0B         0B         0B
```

After modification, you can get the information of meminfo correctly. I am modifying the lookup_initpid_in_store function when considering PR, but this is ambiguous because fc->pid is not initpid. So the modification is modified in multiple xxx_read functions. Are there any other suggestions?

**Comment by hallyn** [comment]
> Hmm, but in that case, your fix would actually lead to showing the wrong data to the container wouldn't it?
> 
> Because a container with `--pid=host` could and likely will still have cgroup limits applied to itself and so lxcfs is still needed to show the proper values in proc files.

Yeah, so would it be better to instead of changing 'if (initpid <= 0)' to 'if (initpid <= 1)', define a helper which checks the pidns of the vicitim task to /proc/self/ns/pid, and sets initpid = fc->pid only if initpid==1 && those pidns inodes are the same?

**Comment by yinhongbo** [comment]
> > Hmm, but in that case, your fix would actually lead to showing the wrong data to the container wouldn't it?
> > Because a container with `--pid=host` could and likely will still have cgroup limits applied to itself and so lxcfs is still needed to show the proper values in proc files.
> 
> Yeah, so would it be better to instead of changing 'if (initpid <= 0)' to 'if (initpid <= 1)', define a helper which checks the pidns of the vicitim task to /proc/self/ns/pid, and sets initpid = fc->pid only if initpid==1 && those pidns inodes are the same?

@hallyn I have modified the code again, please review.

**Comment by yinhongbo** [comment]
ping @Blub

**Comment by yinhongbo** [comment]
Does this PR have new progress? Is it going to do extra work for the merger?

File: bindings.c [hunk 1]
```diff
@@ -662,6 +662,42 @@ static char *slurp_file(const char *from, int fd)
...
+ */
+static int in_same_ns(pid_t pid1, pid_t pid2, char *ns_type)
+{
+    char ns_path1[1024], ns_path2[1024];
+    int s1, s2;
+    struct stat buf1, buf2;
+
+    s1 = snprintf(ns_path1, sizeof(ns_path1)-1, "/proc/%d/ns/%s", pid1, ns_type);
+    if (s1 > 0)
+        ns_path1[s1] = '\0';
```
brauner: Hm, that's not really correct as you're only checking the inode numbers but not the device numbers.
I've added functions in LXC that deal with this correctly. Just copy them over, please:
https://github.com/lxc/lxc/blob/faec88e5ac556d41fb4ee3c12f6fe956070ed516/src/lxc/attach.c#L158
https://github.com/lxc/lxc/blob/345a21ca9ec1b736208611f4bec7e24097ce279b/src/lxc/utils.c#L1379

They may contain a few macros that you might have to get rid off or copy over too and you might need to adapt them a little but I think they are better templates.
Also, if you only use this function to answer 'yes' or 'no', please return a bool.

bindings.c
```diff
@@ -662,6 +662,42 @@ static char *slurp_file(const char *from, int fd)
...
+}
+
+static bool is_host_pid_mode(pid_t pid)
+{
+	if (pid == 1){
+		if (in_same_ns(pid, getpid(), "pid") == 1)
+			return true;
+	}
+	return false;
+}
```
brauner: That's not really our coding style :)
```c
static bool is_host_pid_mode(pid_t pid)
{
	if (pid == 1 && in_same_ns(pid, getpid(), "pid"))
	        return true;

	return false;
}
```

yinhongbo:
@brauner This patch is a good better, thanks. I have applied and submitted

brauner:
> > > Hmm, but in that case, your fix would actually lead to showing the wrong data to the container wouldn't it?
> > > Because a container with `--pid=host` could and likely will still have cgroup limits applied to itself and so lxcfs is still needed to show the proper values in proc files.
> > 
> > 
> > Yeah, so would it be better to instead of changing 'if (initpid <= 0)' to 'if (initpid <= 1)', define a helper which checks the pidns of the vicitim task to /proc/self/ns/pid, and sets initpid = fc->pid only if initpid==1 && those pidns inodes are the same?
> 
> @hallyn I have modified the code again, please review.

So @hallyn's review hasn't been fully addressed, right? Your check stil does:
```c
if (initpid <= 0 || in_same_ns())
```
and not like @hallyn requested
```c
if (initpid <= 1 || in_same_ns())
```

yinhongbo:
> > > > Hmm, but in that case, your fix would actually lead to showing the wrong data to the container wouldn't it?
> > > > Because a container with `--pid=host` could and likely will still have cgroup limits applied to itself and so lxcfs is still needed to show the proper values in proc files.
> > > 
> > > 
> > > Yeah, so would it be better to instead of changing 'if (initpid <= 0)' to 'if (initpid <= 1)', define a helper which checks the pidns of the vicitim task to /proc/self/ns/pid, and sets initpid = fc->pid only if initpid==1 && those pidns inodes are the same?
> > 
> > 
> > @hallyn I have modified the code again, please review.
> 
> So @hallyn's review hasn't been fully addressed, right? Your check stil does:
> 
> ```c
> if (initpid <= 0 || in_same_ns())
> ```
> 
> and not like @hallyn requested
> 
> ```c
> if (initpid <= 1 || in_same_ns())
> ```

has been modified

brauner:
I think this is fine now.

brauner:
Thanks for the patch and sorry for the delay. I've lost track of it.

Issue #125, opened 2016-08-09
ps ux always returns  no btime in /proc/stat


giggls:
Hello,

I'm using lxcfs (version 2.0.2 from debian-backports) on a Debian stable host running a vanilla Kernel (v 4.6.5).

Unfortunately I get "no btime in /proc/stat" inside **all** of my containers (Debian 7.x and 8.x).

I extracted the code in Question from procps sources, because btime **does** actually exist in /proc/stat (verified using `grep btime /proc/stat.`).

Digging further I found that fgets (see the code below) gives me zero bytes after reading "intr ..." which quits the while loop before btime has actually been reached.

So the bug in lxfs seems to be, that ps assumes, that fgets will never return zero until EOF of /proc/stat is reached. Unfortunately lxfs does not seem to make shure, that this assumption is always true.

Here is the code snippet from ps I have been using to debug this:

```
#include <stdio.h>
#include <stdlib.h>
#include <stdbool.h>
#include <string.h>
#define STAT_FILE "/proc/stat"


static char buf[8192];
char buf2[100];

unsigned long getbtime(void) {
    static unsigned long btime = 0;
    bool found_btime = false;
    FILE *f;

    if (btime)
        return btime;

    /* /proc/stat can get very large on multi-CPU systems so we
       can't use FILE_TO_BUF */
    if (!(f = fopen(STAT_FILE, "r"))) {
        fputs("can not open"STAT_FILE, stderr);
        fflush(NULL);
        exit(102);
    }
    /* gets seems to give 0 here in some cases inside containers
       before EOF is reached */ 
    while ((fgets(buf, sizeof buf, f))) {
        if (sscanf(buf, "btime %lu", &btime) == 1) {
            found_btime = true;
            break;
        }
    }
    fclose(f);

    if (!found_btime) {
        fputs("missing btime in " STAT_FILE "\n", stderr);
        exit(1);
    }

    return btime;
}

int main() {
  printf("%ld\n",getbtime());
  exit(0);
}
```

hallyn:
I can't reproduce this on 4.4 and older kernels, don't have a 4.6 kernel handy.  Can you show me the host's and container's /proc/stat contents?

hallyn:
I also cannot reproduce this on Linux x1 4.6.0-10-generic #12-Ubuntu SMP Thu Aug 4 20:52:56 UTC 2016 x86_64 x86_64 x86_64 GNU/Linux

giggls:
[...]

PR #309, opened 2019-10-15
fix read_file second reading empty bug

yinhongbo:
This PR fix is similar to #125 

This type of problem is mainly due to a bug in the read_file function. The reason for this bug will be analyzed below.
Under some Case, We execute ps aux in the container and we get an error in not finding btime.
```
58eb38977abb(@:):/# ps aux
USER         PID %CPU %MEM    VSZ   RSS TTY      STAT START   TIME COMMAND
missing btime in /proc/stat
root           1  0.0  0.0 205172  7412 ?        Ss58eb38977abb(@:):/#
58eb38977abb(@:):/#
```
Then we use strace to track system calls.
```
58eb38977abb(@:):/# strace -ff -s 10240 -e read,open,openat ps aux > /dev/null
...  ...
open("/proc/stat", O_RDONLY)            = 6
read(6, "cpu  13030938 7115021 44303748 6188805307 13413081 0 182664 0 0 0\ncpu0 333372 172944 1060094 154237893 267468 0 13827 0 0 0\ncpu1 414513 177557 1145335 154878583 27030 0 13029 0 0 0\ncpu2 388653 177135 1152247 154818449 48530 0 12159 0 0 0\ncpu3 403940 ... ... 0 0 0 0 0 0 0 0", 4096) = 4096
read(6, "", 4096)                       = 0
read(6, "", 4096)                       = 0
missing btime in /proc/stat
+++ exited with 1 +++
58eb38977abb(@:):/#
```
Then we found that it was normal using cat /proc/stat.
```
58eb38977abb(@:):/# wc /proc/stat
48 2035 5899 /proc/stat
```
We found that the length was 5899 by cat /proc/stat. then we found that the first read got 4096 bytes, and the second read read was 0. Obviously there is a problem with the second read here.

So, I started debugging the lxcfs stat_read.
```
    char *cache = d->buf + CPUALL_MAX_SIZE;
    size_t cache_size = d->buflen - CPUALL_MAX_SIZE;
    FILE *f = NULL;
    struct cpuacct_usage *cg_cpu_usage = NULL;
    int cg_cpu_usage_size = 0;

    if (offset){
        if (offset > d->size)
            return -EINVAL;
        if (!d->cached)
            return 0;
        int left = d->size - offset;
        total_len = left > size ? size: left;
        memcpy(buf, d->buf + offset, total_len);
        return total_len;
    }

    pid_t initpid = lookup_initpid_in_store(fc->pid);
    lxcfs_v("initpid: %d\n", initpid);
    if (initpid <= 0)
        initpid = fc->pid;
```
I found that the offset value of the second read is 4096, which is in line with expectations, but it is true when judging `if (!d->cached)`, and returns directly to 0. Obviously this is not as expected. The problem is that the d->cached is not processed after the first read. Then fix it in the read_file function. 
When repaired, the `ps aux` command is work.

Please review. Thanks.

### DIFF
--- a/bindings.c
+++ b/bindings.c
@@ -3382,6 +3382,8 @@ int read_file(const char *path, char *buf, size_t size, struct file_info *d)
   err:
 	fclose(f);
 	free(line);
+	if (d->size > rv)
+		d->cached = d->size - rv;
 	return rv;
 }
 


brauner:
Thanks for the patch!

PR #314, opened 2020-01-17
add some metrics for meminfo

wlm86:
dirty/writeback  ... and so on

Signed-off-by LiMing Wu <19092205@suning.com>

### DIFF
--- a/bindings.c
+++ b/bindings.c
@@ -63,6 +63,28 @@ struct cpuacct_usage {
 	bool online;
 };
 
+struct memory_stat {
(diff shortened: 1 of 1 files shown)


brauner:
Thanks for the patch. Both clang and gcc are currently unhappy unfortunately.

brauner:
If you fixed those bugs please ping me since I won't get notified when you simply force-push I think.

wlm86:
> If you fixed those bugs please ping me since I won't get notified when you simply force-push I think.

thx, I have fixed

PR #316, opened 2020-01-24
handle swap reporting better when swap accounting is enabled

KellenRenshaw:
Changed the logic used for SwapFree when swap accounting is enabled to better handle situations where memswusage is less than memusage, caused by the fuzziness of the usage_in_bytes counters used as the source.

Signed-off-by: Kellen Renshaw <kellen.renshaw@canonical.com>


KellenRenshaw:
Redid the commit to remove a change that I made and then reverted.

brauner:
Thanks!

stgraber:
@brauner can you make sure it's in stable-3.0? Had a direct request for that.

brauner:
On March 3, 2020 6:45:20 PM GMT+01:00, "Stéphane Graber" <notifications@github.com> wrote:
>@brauner can you make sure it's in stable-3.0? Had a direct request for
>that.

Done.

PR #337, opened 2020-03-03
meminfo: fix used/cache memory calculation under cgroup2

irnes:
Signed-off-by: Irnes Mujkanovic <irnes.mujkanovic@gmail.com>


brauner:
Thanks!

PR #361, opened 2020-03-07
Unified cgroup fixups

3XX0:
Implementation is debatable and minimal, but this was required to have at least `/proc/stat`, `/proc/cpuinfo`, `/proc/meminfo`, and `/sys/devices/system/cpu/online` behave under a unified layout.

### DIFF
--- a/src/cgroups/cgfsng.c
+++ b/src/cgroups/cgfsng.c
@@ -768,7 +768,7 @@ static bool cgfsng_can_use_cpuview(struct cgroup_ops *ops)
 	struct hierarchy *cpu, *cpuacct;
 
 	if (pure_unified_layout(ops))
-		return false;
+		return true;
 
 	cpu = ops->get_hierarchy(ops, "cpu");
 	if (!cpu || is_unified_hierarchy(cpu))
--- a/src/proc_cpuview.c
+++ b/src/proc_cpuview.c
@@ -421,13 +421,19 @@ static bool read_cpu_cfs_param(const char *cg, const char *param, int64_t *value
 {
 	__do_free char *str = NULL;
 	char file[11 + 6 + 1]; /* cpu.cfs__us + quota/period + \0 */
+	bool first = true;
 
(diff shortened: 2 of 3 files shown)


brauner:
This is the first time the test-suite is fully up and running. :D Kinda excited to see whether it survives your patch. :D

brauner:
Wohoo, passed with flying colors. :)

brauner:
Thanks!

brauner:
> Implementation is debatable and minimal, but this was required to have at least `/proc/stat`, `/proc/cpuinfo`, `/proc/meminfo`, and `/sys/devices/system/cpu/online` behave under a unified layout.

It's fine we've had to hack around various restrictions for quite a while until cgroup v1 was even ready to support virtualizing these files.

3XX0:
For the record, we might want to clean it up later (maybe unify both `read_cpu_cfs_param` calls) and add the `cpuacct` equivalent of cgroupv2.

I smoke tested it with `systemd-run --scope --user -p AllowedCPUs=0 -p CPUQuota=50% -p MemoryMax=500M ...`

brauner:
Yeah, we should probably port this to dedicated methods on the cgroup struct in cgroups/.

3XX0:
Um actually the patch for meminfo is only partially correct...
The `retlimit = 0` only works for swap when it's off.
Cgroupv2 puts the string `max` everywhere when there is no limit or no swap, so the actual fix is more complicated.

brauner:
On March 7, 2020 12:34:31 PM GMT+01:00, Jonathan Calmels <notifications@github.com> wrote:
>Um actually the patch for meminfo is only partially correct...
>The `retlimit = 0` only works for swap when it's off.
>Cgroupv2 puts the string `max` everywhere when there is no limit or no
>swap, so the actual fix is more complicated.

Want this to be reverted or want to send further fixes?

3XX0:
I will try to figure out a way to fix it, but this is still better than before so we might as well keep this patch

PR #368, opened 2020-03-14
tree-wide: use {u}int64_t types

brauner:


stgraber:
@brauner any risk on upgrades from older version using the old types?

brauner:
There should'nt be any, as the whole library gets reloaded so the old types can't be accessed anymore.

PR #396, opened 2020-04-17
cgroups: more cgroup2 fun

brauner:
Try too read a valid value from a given cgroup file. If it is a legacy
cgroup hierarchy and we fail to find a valid value we terminate early
and report an error.
The cgroup2 hierarchy however, has different semantics. In a few
controller files it will show the value "max" or simply leave it
completely empty thereby indicating that no limit has been set for this
particular cgroup.  However, that doesn't mean that there's no limit. A
cgroup further up the hierarchy could have a limit set that also applies
to the cgroup we are interested in. So for the unified cgroup hierarchy
we need to keep walking towards the cgroup2 root cgroup and try to parse
a valid value.

### DIFF
--- a/src/cgroups/cgroup_utils.c
+++ b/src/cgroups/cgroup_utils.c
@@ -724,3 +724,102 @@ bool mkdir_p(const char *dir, mode_t mode)
 
 	return true;
 }
+
+static bool same_file(int fd1, int fd2)
+{
+	struct stat st1, st2;
+
+	if (fstat(fd1, &st1) < 0 || fstat(fd2, &st2) < 0)
+		return false;
+
+	return (st1.st_dev == st2.st_dev) && (st1.st_ino == st2.st_ino);
+}
+
+/**
+ * cgroup_walkup_to_root() - Walk upwards to cgroup root to find valid value
+ *
+ * @cgroup2_root_fd:	File descriptor for the cgroup2 root mount point.
+ * @hierarchy_fd:	File descriptor for the hierarchy.
+ * @cgroup:		A cgroup directory relative to @hierarchy_fd.
+ * @file:		The file in @cgroup from which to read a value.
+ * @value:		Return argument to store value read from @file.
+ *
+ * This function tries to read a valid value from @file in @cgroup in
+ * @hierarchy_fd. If it is a legacy cgroup hierarchy and we fail to find a
+ * valid value we terminate early and report an error.
+ * The cgroup2 hierarchy however, has different semantics. In a few controller
+ * files it will show the value "max" or simply leave it completely empty
+ * thereby indicating that no limit has been set for this particular cgroup.
+ * However, that doesn't mean that there's no limit. A cgroup further up the
+ * hierarchy could have a limit set that also applies to the cgroup we are
+ * interested in. So for the unified cgroup hierarchy we need to keep walking
+ * towards the cgroup2 root cgroup and try to parse a valid value.
+ */
+int cgroup_walkup_to_root(int cgroup2_root_fd, int hierarchy_fd,
+			  const char *cgroup, const char *file, char **value)
+{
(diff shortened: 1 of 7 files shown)


src/cgroups/cgroup_utils.c
```diff
@@ -724,3 +724,109 @@ bool mkdir_p(const char *dir, mode_t mode)
...
+		}
+
+		/*
+		 * Set an arbitraty hard-coded limit to prevent us from ending
+		 * up in an endless loop. There really shouldn't be any cgroup
+		 * tree that is 1000 levels deep. That would be insane in
+		 * principal and performance-wise.
+		 */
+		if (i == 1000)
+			return log_error_errno(-ELOOP, ELOOP, "To many nested cgroups or invalid mount tree. Terminating walk");
```
Blub, line 823: Maybe put the condition at the top of the loop and not have a `not reached` part below? :-)

src/cgroups/cgroup_utils.c
```diff
@@ -724,3 +724,109 @@ bool mkdir_p(const char *dir, mode_t mode)
...
+		 */
+		if (same_file(cgroup2_root_fd, cur_fd)) {
+			if (no_limit)
+				return 1;
+
+			return -EINVAL;
+		}
+
+		/* We found a valid value. Terminate walk. */
+		new_val = readat_file(new_dir_fd, file);
```
Blub, line 808: Shouldn't this already read from `cur_fd`? Otherwise the first iteration reads the same file as we read right on top of the function.

src/cgroups/cgroup_utils.c
brauner: Sounds good. Fixed.

src/cgroups/cgroup_utils.c
```diff
@@ -724,3 +724,109 @@ bool mkdir_p(const char *dir, mode_t mode)
...
+		 * up in an endless loop. There really shouldn't be any cgroup
+		 * tree that is 1000 levels deep. That would be insane in
+		 * principal and performance-wise.
+		 */
+		if (i == 1000)
+			return log_error_errno(-ELOOP, ELOOP, "To many nested cgroups or invalid mount tree. Terminating walk");
+
+		swap_fd = move_fd(cur_fd);
+		cur_fd = move_fd(new_dir_fd);
+		new_dir_fd = swap_fd;
```
Blub, line 827: Pretty sure one of these should be using `dir_fd`, since both `cur_fd` and `new_dir_fd` are inside this scope and therefore get closed at this point I think.

brauner: dir_fd is moved before the first openat()

src/cgroups/cgroup_utils.c
brauner: Yep, fixed!

src/cgroups/cgroup_utils.c
brauner: oh right, i see

brauner: Check whether what I have now makes sense, please?

src/cgroups/cgroup_utils.c
```diff
@@ -724,3 +724,100 @@ bool mkdir_p(const char *dir, mode_t mode)
...
+				return 1;
+
+			return -EINVAL;
+		}
+
+		/* We found a valid value. Terminate walk. */
+		new_val = readat_file(dir_fd, file);
+		if (is_empty_string(new_val) || strcmp(new_val, "max") == 0) {
+			no_limit = true;
+		} else if (!is_empty_string(new_val)) {
```
Blub, line 816: I think this could just be `} else {`, since `!is_empty_string` bust be the case otherwise we'd be in the upper branch?

brauner: Will fix.

PR #400, opened 2020-04-24
proc_fuse: silence error when we find no memlimit

Blub:
We get a positive result with an empty string if we do not
encounter any limits while walking up the cgroup.


brauner:
Thanks!

PR #426, opened 2020-09-03
Add support for fuse3

sparlane:
Signed-off-by: Scott Parlane <scott.parlane@alliedtelesis.co.nz>

### DIFF
--- a/src/bindings.c
+++ b/src/bindings.c
@@ -4,8 +4,16 @@
 #define _GNU_SOURCE
 #endif
 
+#include "config.h"
+
+#ifdef HAVE_FUSE3
+#ifndef FUSE_USE_VERSION
+#define FUSE_USE_VERSION 30
+#endif
+#else
 #ifndef FUSE_USE_VERSION
 #define FUSE_USE_VERSION 26
+#endif
 #endif
 
 #define _FILE_OFFSET_BITS 64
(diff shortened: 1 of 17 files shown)


sparlane:
This is currently a WIP, it builds, but I haven't tested it yet.

brauner:
> This is currently a WIP, it builds, but I haven't tested it yet.

Thanks for working on this! Please ping me when it is in reviewable shape!

sparlane:
@brauner 
Manual testing with it built against fuse-3.9.2 indicates it now works as expected.

src/bindings.c
```diff
@@ -4,9 +4,17 @@
...
+
+#ifdef HAVE_FUSE3
+#ifndef FUSE_USE_VERSION
+#define FUSE_USE_VERSION 30
+#endif
+#else
 #ifndef FUSE_USE_VERSION
 #define FUSE_USE_VERSION 26
 #endif
+#endif
```
brauner, line 17: I wonder if we shouldn't just encapsulate this in a common header `fuse_version.h` instead of duplicating this in every file.

brauner: Similar to what you already did with `fuse_compat.h` below.

brauner (approved):
One small request for the future but overall this looks sane to me!

brauner:
Thank you very much for this work!

PR #428, opened 2020-09-03
Fix `get_min_memlimit()` on non-glibc

devplayer0:
`dirname()` behaves differently on `glibc` vs POSIX implementations (e.g. `musl`).

Under `glibc`, the use of `dirname()` terminates at `/` as required, but the same code on `musl` ends at, for example, `/lxc.payload.dear-raccoon`, resulting in an infinite loop when trying to read `/proc/meminfo` in a container.

A simple replacement for `dirname()` solves this on non-`glibc` platforms like Alpine. :smile:

### DIFF
--- a/src/proc_fuse.c
+++ b/src/proc_fuse.c
@@ -213,6 +213,68 @@ static uint64_t get_memlimit(const char *cgroup, bool swap)
 	return memlimit;
 }
 
+/*
+ * This function taken from glibc-2.32, as POSIX dirname("/some-dir") will
+ * return "/some-dir" as opposed to "/", which breaks `get_min_memlimit()`
+ */
+static char *gnu_dirname(char *path)
+{
+	static const char dot[] = ".";
+	char *last_slash;
+
+	/* Find last '/'.  */
+	last_slash = path != NULL ? strrchr(path, '/') : NULL;
+
+	if (last_slash != NULL && last_slash != path && last_slash[1] == '\0') {
+		/* Determine whether all remaining characters are slashes.  */
+		char *runp;
+
+		for (runp = last_slash; runp != path; --runp)
+			if (runp[-1] != '/')
+				break;
+
+		/* The '/' is the last character, we have to look further.  */
+		if (runp != path)
+			last_slash = memrchr(path, '/', runp - path);
+	}
+
+	if (last_slash != NULL) {
+		/* Determine whether all remaining characters are slashes.  */
+		char *runp;
+
+		for (runp = last_slash; runp != path; --runp)
+			if (runp[-1] != '/')
+				break;
+
+		/* Terminate the path.  */
+		if (runp == path) {
+			/*
+			 * The last slash is the first character in the string.
+			 * We have to return "/".  As a special case we have to
+			 * return "//" if there are exactly two slashes at the
+			 * beginning of the string.  See XBD 4.10 Path Name
+			 * Resolution for more information
+			 */
+			if (last_slash == path + 1)
+				++last_slash;
+			else
+				last_slash = path + 1;
+		} else
+			last_slash = runp;
+
+		last_slash[0] = '\0';
+	} else {
+		/*
+		 * This assignment is ill-designed but the XPG specs require to
+		 * return a string containing "." in any case no directory part
+		 * is found and so a static and constant string is required.
+		 */
+		path = (char *)dot;
+	}
+
+	return path;
+}
+
 static uint64_t get_min_memlimit(const char *cgroup, bool swap)
 {
 	__do_free char *copy = NULL;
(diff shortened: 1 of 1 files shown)


src/proc_fuse.c
```diff
@@ -213,6 +213,25 @@ static uint64_t get_memlimit(const char *cgroup, bool swap)
...
 
+/*
+ * dirname() behaves differently on musl than glibc, so implement the bits we
+ * need (the path in this case is always absolute and never ends in a `/`)
+ */
+static char *cgroup_parent(char *cgroup)
+{
+	if (strcmp(cgroup, "/") == 0) {
+		return cgroup;
+	}
```
brauner, line 224: We don't use {} around single-line if statements. :)

src/proc_fuse.c
```diff
@@ -213,6 +213,25 @@ static uint64_t get_memlimit(const char *cgroup, bool swap)
...
+
+	char *last_slash = strrchr(cgroup, '/');
+	if (last_slash == cgroup) {
+		last_slash++;
```
brauner, line 228: We don't use {} around single-line if statements. :)

src/proc_fuse.c
```diff
@@ -213,6 +213,25 @@ static uint64_t get_memlimit(const char *cgroup, bool swap)
...
+ * dirname() behaves differently on musl than glibc, so implement the bits we
+ * need (the path in this case is always absolute and never ends in a `/`)
+ */
+static char *cgroup_parent(char *cgroup)
+{
+	if (strcmp(cgroup, "/") == 0) {
+		return cgroup;
+	}
+
+	char *last_slash = strrchr(cgroup, '/');
```
brauner, line 226: Please declare variables at the beginning of a scope. :)

src/proc_fuse.c
```diff
@@ -213,6 +213,25 @@ static uint64_t get_memlimit(const char *cgroup, bool swap)
 	return memlimit;
 }
 
+/*
+ * dirname() behaves differently on musl than glibc, so implement the bits we
+ * need (the path in this case is always absolute and never ends in a `/`)
+ */
+static char *cgroup_parent(char *cgroup)
+{
+	if (strcmp(cgroup, "/") == 0) {
```
brauner, line 222: This can just be *cgroup == '/'

brauner (changes requested):
Fun. Thanks for the patch. :)

devplayer0:
Thanks for the comments, updated :)

src/proc_fuse.c
devplayer0: Oops, actually just realised it has to be `strcmp()`, `*cgroup == '/'` will always be true haha.

src/proc_fuse.c
```diff
@@ -213,6 +213,24 @@ static uint64_t get_memlimit(const char *cgroup, bool swap)
...
+	if (strcmp(cgroup, "/") == 0)
+		return cgroup;
+
+	last_slash = strrchr(cgroup, '/');
+	if (last_slash == cgroup)
+		last_slash++;
+	*last_slash = '\0';
+
+	return cgroup;
+}
```
brauner, line 232: Hm, I think this doesn't really work. What do you expect the output of:
"foo/bar" to be for `cgroup_parent()` vs `dirname()`?
`dirname(foo/bar)` gives you "foo" but `cgroup_parent()` gives you "foo/bar".

devplayer0: Actually I think `cgroup_parent()` does give you `"foo"` in that case? Where `cgroup_parent()` breaks down (most critically, there are definitely more edge cases that GNU `dirname()` covers) is with `cgroup_parent("foo")`, which segfaults. I could add a check for this but in the use case of `get_min_memlimit()` (as far as I can tell) the path will always start with a `/`. The `strcmp(copy, "/") == 0` in `gen_min_memlimit()` would break in the case of a path like `"foo"` anyway?

Am I right in thinking the value is only ever the third item in a `/proc/*/cgroup` tuple (e.g. `/user.slice/user-1000.slice/session-1.scope`?)

brauner:
On Fri, Sep 04, 2020 at 03:33:31AM -0700, devplayer0 wrote:
> @devplayer0 commented on this pull request.
> 
> 
> 
> > + * dirname() behaves differently on musl than glibc, so implement the bits we
> + * need (the path in this case is always absolute and never ends in a `/`)
> + */
> +static char *cgroup_parent(char *cgroup)
> +{
> +	char *last_slash = NULL;
> +	if (strcmp(cgroup, "/") == 0)
> +		return cgroup;
> +
> +	last_slash = strrchr(cgroup, '/');
> +	if (last_slash == cgroup)
> +		last_slash++;
> +	*last_slash = '\0';
> +
> +	return cgroup;
> +}
> 
> Actually I think `cgroup_parent()` does give you `"foo"` in that case? Where `cgroup_parent()` breaks down (most critically, there are definitely more edge cases that GNU `dirname()` covers) is with `cgroup_parent("foo")`, which segfaults. I could add a check for this but in the use case of `get_min_memlimit()` (as far as I can tell) the path will always start with a `/`. The `strcmp(copy, "/") == 0` in `gen_min_memlimit()` would break in the case of a path like `"foo"` anyway?
> 
> Am I right in thinking the value is only ever the third item in a `/proc/*/cgroup` tuple (e.g. `/user.slice/user-1000.slice/session-1.scope`?)

Yeah, it should be. But I don't want to rely us on this being sane. When
cgroup2 got introduced we already ran into issues because there suddenly
was a line that had a different format ("0::/foo/bar").

(One thing you could consider is to copy dirname from misc/dirname.c in
glibc and add it to lxcfs as lxcfs_dirname() adding a comment where
you've taken it from. Licenses should be compatible.)

Christian

devplayer0:
Cool, I was originally looking at the `glibc` implementation anyway, so I'll paste that in with a comment. :) I'll update `get_min_memlimit()` so that it won't fail with a path that doesn't start with a `/` too.

brauner (approved):

PR #436, opened 2020-11-05
README: clarify SWAP

stgraber:
Signed-off-by: Stéphane Graber <stgraber@ubuntu.com>

### DIFF
--- a/README.md
+++ b/README.md
@@ -104,3 +104,50 @@ docker run -it -m 256m --memory-swap 256m \
  In a system with swap enabled, the parameter "-u" can be used to set all values in "meminfo" that refer to the swap to 0.
 
  sudo lxcfs -u /var/lib/lxcfs
+
+## Swap handling
+If you noticed LXCFS not showing any SWAP in your container despite
+having SWAP on your system, please read this section carefully and look
+for instructions on how to enable SWAP accounting for your distribution.
+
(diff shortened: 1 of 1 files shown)


stgraber:
@brauner anything else I should cover in there?

If not, then we need to make sure that LXCFS behaves as described. Currently in testing, I've noticed that:
 - We don't do the proper SWAP device size when a memory limit is applied
 - We certainly don't have that swappinness logic I've added in there, though I think it would make sense and make the disable-swap knob we have in LXD be more visible

brauner:
> @brauner anything else I should cover in there?
> 
> If not, then we need to make sure that LXCFS behaves as described. Currently in testing, I've noticed that:
> 
> * We don't do the proper SWAP device size when a memory limit is applied

What do you see?

> * We certainly don't have that swappinness logic I've added in there, though I think it would make sense and make the disable-swap knob we have in LXD be more visible

Do we really need more complications in that code? Has anyone asked for this feature? Sure, we can do that but it's one more knob to turn and this doesn't seem to be supported in cgroup2.

stgraber:
Updated the entry to act more as an entry point for anyone missing SWAP in LXCFS.
Also reworded things a bit to not reference swapaccount too much.

brauner:
Thank you!

Issue #444, opened 2021-02-03
"ps" show 0.0 of %cpu for a busy process in container


borgerli:
I'm using lxcfs 4.0.7.  I created an container whit lxcfs proc files mount, then kicked off a process `while true;do echo test > /dev/null;done`. In container, `top` command showed the correct %cpu information, while `ps` always showed  `0.0`. However when not using lxcfs, `ps` worked well.

### start lxcfs
`/usr/local/bin/lxcfs -l --enable-cfs --enable-pidfd /var/lib/lxc/lxcfs

### start docker container
```
docker run -it -m 128m --cpus=1 --rm \
  -v /var/lib/lxc/lxcfs/proc/cpuinfo:/proc/cpuinfo:rw \
  -v /var/lib/lxc/lxcfs/proc/diskstats:/proc/diskstats:rw \
  -v /var/lib/lxc/lxcfs/proc/meminfo:/proc/meminfo:rw \
  -v /var/lib/lxc/lxcfs/proc/stat:/proc/stat:rw \
  -v /var/lib/lxc/lxcfs/proc/swaps:/proc/swaps:rw \
  -v /var/lib/lxc/lxcfs/proc/loadavg:/proc/loadavg:rw \
  -v /var/lib/lxc/lxcfs/proc/uptime:/proc/uptime:rw \
  -v /var/lib/lxc/lxcfs/sys/devices/system/cpu/online:/sys/devices/system/cpu/online:rw \
  -v /var/lib/lxc:/var/lib/lxc:rshared \
  centos:7 /bin/bash
```
### test

`top` shows `100.0`, while `ps` shows `0.0` for process 16

```
[root@af61796cf0ed /]# while true; do echo test > /dev/null;done &
[1] 16
[root@af61796cf0ed /]# top -b -n 1
top - 03:06:17 up 0 min,  0 users,  load average: 0.00, 0.00, 0.00
Tasks:   3 total,   2 running,   1 sleeping,   0 stopped,   0 zombie
%Cpu(s): 12.0 us,  0.0 sy,  0.0 ni, 88.0 id,  0.0 wa,  0.0 hi,  0.0 si,  0.0 st
KiB Mem :   131072 total,   127272 free,     3800 used,        0 buff/cache
KiB Swap:        0 total,        0 free,        0 used.   127272 avail Mem 

  PID USER      PR  NI    VIRT    RES    SHR S  %CPU %MEM     TIME+ COMMAND
[...]

PR #445, opened 2021-02-07
fix 0.0 pcpu issue for ps command

borgerli:
fix issue #444 

Signed-off-by: LI Bo borgerli@tencent.com

### DIFF
--- a/src/proc_fuse.c
+++ b/src/proc_fuse.c
@@ -19,6 +19,8 @@
 #define _FILE_OFFSET_BITS 64
 
 #define __STDC_FORMAT_MACROS
+#define TASK_COMM_LEN 16
+#define TASK_COMM_LEN_STR "17"
 #include <dirent.h>
 #include <errno.h>
 #include <fcntl.h>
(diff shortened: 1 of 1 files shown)


brauner:
Oh wow, I didn't get notifed about this at all. I'm sorry.

brauner:
Can you please sign-off you commit? :)

borgerli:
Oh, I see. Thanks

> Can you please sign-off you commit? :)

borgerli:
Signed-off-by: LI Bo borgerli@tencent.com

borgerli:
@brauner, I signed off this PR. Could you please help review?  Thanks.

stgraber:
@brauner can you take a look?

brauner (commented):
I've reviewed this a while ago but I'm not too happy with this approach. The fix here means we're special-casing `ps` which I think is wrong. I would like to understand the root cause of this as this may easily happen with other tools too and we can't keep growing a list of psecial-cased tools. :)

borgerli:
Yes, this PR just fixes a special case. Looking forward to a general solution for such cases. Thanks.

Issue #448, opened 2021-03-12
/proc/cpuinfo and /proc/stat truncated on high-cpu count systems


multippt:
When lxcfs is built using fuse3, the output for /proc/cpuinfo and /proc/stat is truncated to 4096 bytes.
The limit appears to be set by https://github.com/lxc/lxcfs/commit/2071129a317d7553657c94714c9f3fcd5ca4bde2 

This issue affects 4.0.6 or newer on machines that have a large cpu count (e.g. > 4).

PR #449, opened 2021-03-17
Determine the file size from the current system file size

sparlane:
fixes: #448 caused by 25982f5d7e008e6ad13df28af3006ab8286f4c97

Signed-off-by: Scott Parlane <scott.parlane@alliedtelesis.co.nz>

### DIFF
--- a/src/proc_fuse.c
+++ b/src/proc_fuse.c
@@ -80,6 +80,23 @@ struct memory_stat {
 	uint64_t total_unevictable;
 };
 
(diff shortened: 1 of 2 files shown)


sparlane:
We are also sometimes seeing another bug with the file size when it is too big for the amount of data returned, frequency is about 1 in 1000-5000, I will look to address that soon.

brauner:
Looks good, thanks!

PR #465, opened 2021-05-26
sys: virtualize /sys/devices/system/cpu/cpu*

brauner:
Signed-off-by: Zhiwei Li <cyzz_a102@126.com>


brauner:
Hm, I'm beginning to think we need this under a separate command line argument.

brauner:
> Hm, I'm beginning to think we need this under a separate command line argument.

Ok, I think I have an upgradable version that doesn't require a new flag argument.

brauner:
> > Hm, I'm beginning to think we need this under a separate command line argument.
> 
> Ok, I think I have an upgradable version that doesn't require a new flag argument.

This will be more complicated than I thought.

PR #474, opened 2021-09-01
meson: add meson support

brauner:


stgraber:
Looks pretty good, thanks!

Can you split:
 - Code refactoring (I see some defines, includes, chunks, ... moving around)
 - Code changes needed because of stricter build flags
 - Removal of autoconf/automake
 - Addition of meson
 - Porting of CI scripts
 - Updating README and any other build instruction

Also, from our discussion yesterday I thought we wanted to keep it building with 16.04 because of our large remaining userbase there. So we'd need to re-introduce that Github Action chunk and make it behave.

That should make it easier to review as well as backport.
For stable-4.0, I'm tempted to have us offer both. That way we don't regress current build scripts but we also have meson for those who prefer it.

brauner:
> Looks pretty good, thanks!
> 
> Can you split:
> 
> * Code refactoring (I see some defines, includes, chunks, ... moving around)
> * Code changes needed because of stricter build flags
> * Removal of autoconf/automake
> * Addition of meson
> * Porting of CI scripts
> * Updating README and any other build instruction
> 
> Also, from our discussion yesterday I thought we wanted to keep it building with 16.04 because of our large remaining userbase there. So we'd need to re-introduce that Github Action chunk and make it behave.
> 
> That should make it easier to review as well as backport.
> For stable-4.0, I'm tempted to have us offer both. That way we don't regress current build scripts but we also have meson for those who prefer it.

We shouldn't backport this to stable. This is just an invitation for breakage and the test scripts can't support both build systems and I'm very unenthusiastic about hacking them to do so as this is guaranteed to cause mayhem and breakage.

brauner:
h

> Looks pretty good, thanks!
> 
> Can you split:
> 
> * Code refactoring (I see some defines, includes, chunks, ... moving around)
> * Code changes needed because of stricter build flags
> * Removal of autoconf/automake
> * Addition of meson
> * Porting of CI scripts

These steps can't really be switched since this will leave LXCFS in a broken state in between where it can't be compiled. We need to switch in one go. So we should switch to Meson and port the scripts in one commit. Then remove the old build system.

meson.build
```diff
@@ -0,0 +1,248 @@
...
+endforeach
+
+conf.set('_FILE_OFFSET_BITS', 64)
+
+libfuse = cc.find_library('fuse3', required : false)
+if libfuse.found()
+	conf.set10('HAVE_FUSE3', libfuse.found())
+	conf.set('FUSE_USE_VERSION', 30)
+else
+	libfuse = cc.find_library('fuse')
```
eli-schwartz, line 134: Fuse and fuse3 have pkg-config files, fuse.pc and fuse3.pc

I'm not really sure what/how fuse works, but FUSE_USE_VERSION seems like an API level for which the ability to do

```
dependency('fuse', version: '>= 2.6')
```

May be useful?

meson.build
```diff
@@ -0,0 +1,248 @@
...
+	conf.set10('HAVE_FUSE', libfuse.found())
```
eli-schwartz, line 135: It doesn't seem this define is ever used, it is just detected by virtue of the "else" clause in variious HAVE_FUSE3 checks.

meson.build
```diff
@@ -0,0 +1,248 @@
...
+	'src/sysfs_fuse.c',
+	'src/sysfs_fuse.h',
+	'src/utils.c',
+	'src/utils.h')
+
+liblxcfs = shared_library(
+        'lxcfs',
+        liblxcfs_sources,
+        link_args : ['-shared'],
+        dependencies : [threads,
```
eli-schwartz, line 199: -shared is already handled by the shared_library() function. Given the dependencies and liblxcfs_sources are shared between two targets, you could use:

```
commondep = declare_dependency(
    sources: liblxcfs_sources,
    dependencies: [threads, libdl, libfuse])
```

and then just have both shared_library() add a dependencies: commondep

meson.build
eli-schwartz: Also https://mesonbuild.com/Wrap-dependency-system-manual.html

You can add a small file at subprojects/fuse3.wrap:

```
[wrap-file]
directory = libfuse-fuse-3.10.5
source_url = https://github.com/libfuse/libfuse/archive/refs/tags/fuse-3.10.5.tar.gz
source_filename = fuse-3.10.5.tar.gz
source_hash = e73f75e58da59a0e333d337c105093c496c0fd7356ef3a5a540f560697c9c4e6

[provide]
fuse3 = libfuse_dep
```

and meson will automatically download and build a copy of fuse3 if it cannot find one on the system.

meson.build
```diff
@@ -0,0 +1,248 @@
...
+
+cc = meson.get_compiler('c')
+meson_build_sh = find_program('tools/meson-build.sh')
+
+want_tests = get_option('tests')
+
+possible_cc_flags = [
+	'-fPIE',
+	'-Wvla',
+	'-std=gnu11',
```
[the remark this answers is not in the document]
brauner: done

meson.build
brauner: Hm, the problem with this is that `dependency('fuse')` doesn't look for `fuse3` when just specifying `fuse`. No big deal. I'll work around it.

meson.build
brauner: I reworked this, thanks!

meson.build
brauner: Thanks, done!

PR #483, opened 2021-10-05
Bugfix: fix lxcfs --version or -v loss

Sn0rt:
before

image

after

```bash
# lxcfs --version
4.0.9
# lxcfs -v
4.0.9
```

Signed-off-by: sn0rt <wangguohao.2009@gmail.com>


src/lxcfs.c
```diff
@@ -1244,6 +1248,9 @@ int main(int argc, char *argv[])
 		case 'u':
 			opts->swap_off = true;
 			break;
+		case 'v':
+			lxcfs_info("%s", VERSION);
+			exit(EXIT_SUCCESS);
```
brauner, line 1253: This won't compile. You need to add sm like:
```diff
diff --git a/src/lxcfs.c b/src/lxcfs.c
index 172140b..a84b7a9 100644
--- a/src/lxcfs.c
+++ b/src/lxcfs.c
@@ -1244,6 +1244,9 @@ int main(int argc, char *argv[])
                case 'u':
                        opts->swap_off = true;
                        break;
+               case 'v':
+                       lxcfs_info(STRINGIFY(PROJECT_VERSION));
+                       exit(EXIT_SUCCESS);
                default:
                        usage();
                }
diff --git a/src/macro.h b/src/macro.h
index a7fe05d..b9abf11 100644
--- a/src/macro.h
+++ b/src/macro.h
@@ -145,4 +145,7 @@
 #define __returns_twice __attribute__((returns_twice))
 #endif

+#define STRINGIFY(a) __STRINGIFY(a)
+#define __STRINGIFY(a) #a
+
 #endif /* __LXCFS_MACRO_H */
```

meson.build
```diff
@@ -1,7 +1,7 @@
 # SPDX-License-Identifier: LGPL-2.1-or-later
 
 project('lxcfs', 'c',
-        version : '4.0.0',
+        version : '4.0.9',
```
stgraber, line 4: Don't do this, the `4.0.9` version should only be in the `stable-4.0` branch. Master remains at `4.0.0` until we release a `4.1.0` or a `5.0.0`.

src/lxcfs.c
```diff
@@ -1254,6 +1261,9 @@ int main(int argc, char *argv[])
 
 	new_argv = &argv[optind];
 	new_argc = argc - optind;
+	if (new_argc == 0) {
+		usage();
+	}
```
brauner, line 1266: Doesn't seem like that would be needed and I think the order might be better sm like:
```c
if (new_argc == 0) {
        lxcfs_error("Missing mountpoint");
	goto out;
}

if (new_argc > 1)
        usage();
```

brauner (changes requested):

PR #485, opened 2021-10-09
Fix padding for Slab entry in /proc/meminfo

mtstickney:
The padding used for the 'Slab' entry of /proc/meminfo is three bytes
shorter than the padding used by the kernel. Since we report the host's
file size as the size of the lxcfs file, this means reads would be three
bytes short of the reported file length. In some cases this causes the
client to receive three additional NUL bytes at the end of the file,
which can cause various problems.

Signed-off-by: Matthew Stickney <fiendishlinux@gmail.com>


brauner:
Thanks!

PR #507, opened 2021-12-07
lxcfs: massage locking

brauner:

### DIFF
--- a/src/lxcfs.c
+++ b/src/lxcfs.c
@@ -29,6 +29,7 @@
 #include "bindings.h"
 #include "lxcfs_fuse_compat.h"
 #include "macro.h"
+#include "atomic_utils.h"
 #include "memory_utils.h"
 
 void *dlopen_handle;
@@ -37,6 +38,7 @@ void *dlopen_handle;
 
 static int users_count;
 static pthread_mutex_t user_count_mutex = PTHREAD_MUTEX_INITIALIZER;
+static pthread_cond_t cond_zero = PTHREAD_COND_INITIALIZER;
 static void lock_mutex(pthread_mutex_t *l)
 {
 	int ret;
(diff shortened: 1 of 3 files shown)


src/lxcfs.c
```diff
@@ -157,18 +159,26 @@ static void do_reload(void)
...
 {
+	int nr_users;
+
+	nr_users = atomic_fetch_inc(&users_count);
+	if (!need_reload)
+		return;
+
 	users_lock();
-	if (users_count == 0 && need_reload)
+	if (nr_users == 0) {
```
Blub, line 169: I think the `nr_users == 0` check might fail if there were ongoing requests when the first thread after `need_reload` became `true` enters this.

I think we need to decrement after the first `if` again and introduce a second condvar guarding the `do_reload()` which gets triggered in `down_user` if the count drops to 0 and `need_reload` is set?

src/lxcfs.c
```diff
@@ -157,18 +159,26 @@ static void do_reload(void)
...
 		do_reload();
-	users_count++;
+                pthread_cond_signal(&var_cond);
```
Blub, line 171: I think this should be `pthread_cond_broadcast()` to wake up *all* waiters.

src/lxcfs.c
brauner: Hm, `nr_users` is retrieved via `atomic_fetch_inc()` so it shouldn't change anymore after we've retrieved it.

src/lxcfs.c
```diff
@@ -149,31 +151,41 @@ static void do_reload(void)
...
 }
 
 static void up_users(void)
 {
-	users_lock();
-	if (users_count == 0 && need_reload)
-		do_reload();
-	users_count++;
-	users_unlock();
+	if (atomic_load_acquire(&need_reload)) {
```
Blub, line 158: Doing this before the `increment` can be dangerous, though.
Assume thread 1 goes to the *top* of the `else` branch before incrementing, `user_count` is still `0`.
Then `need_reload` gets set and another thread goes into this branch right up to the `do_reload()`, and right in the middle of it, the scheduler switches back to thread 1 which bumps the counter and continues towards a crash.

brauner: Yeah, I already had deadlocks anyway.

PR #508, opened 2021-12-09
make meminfo and swaps cgroupv2 aware

Blub:
Also deduplicates some common code there.

An alternative to the cgv2 check inside `proc_fuse.c` would be to move the mem/swap/memsw-sum logic into the cgroup ops, not sure if it's worth it.


brauner:
Thanks! That looks like a good solution.

PR #524, opened 2022-03-12
meson: Include documentation

Foxboron:
Documentation was removed from the build system with the migration to
meson. This implements the help2man generation which existed in the
autoconf setup.

Signed-off-by: Morten Linderud <morten@linderud.pw>


stgraber:
Thanks!

Issue #522, opened 2022-03-12
/sys/devices/system/cpu virtualization breaks _NPROCESSORS_CONF


stgraber:
```
root@c1:~# getconf _NPROCESSORS_ONLN
4
root@c1:~# getconf _NPROCESSORS_CONF
0
```

Filesystem looks correct:
```
root@c1:~# ls -l /sys/devices/system/cpu/
total 0
drwxr-xr-x 2 nobody nogroup   0 Mar 12 06:35 cpu0
drwxr-xr-x 2 nobody nogroup   0 Mar 12 06:35 cpu1
drwxr-xr-x 2 nobody nogroup   0 Mar 12 06:35 cpu2
drwxr-xr-x 2 nobody nogroup   0 Mar 12 06:35 cpu3
drwxr-xr-x 2 nobody nogroup   0 Mar 12 06:35 cpufreq
drwxr-xr-x 2 nobody nogroup   0 Mar 12 06:35 cpuidle
drwxr-xr-x 2 nobody nogroup   0 Mar 12 06:35 hotplug
-r--r--r-- 1 nobody nogroup   1 Mar 12 06:35 isolated
-r--r--r-- 1 nobody nogroup   5 Mar 12 06:35 kernel_max
-r--r--r-- 1 nobody nogroup 724 Mar 12 06:35 modalias
-r--r--r-- 1 nobody nogroup   1 Mar 12 06:35 offline
-r--r--r-- 1 nobody nogroup   4 Mar 12 06:35 online
-r--r--r-- 1 nobody nogroup   4 Mar 12 06:35 possible
drwxr-xr-x 2 nobody nogroup   0 Mar 12 06:35 power
-r--r--r-- 1 nobody nogroup   4 Mar 12 06:35 present
-rw-r--r-- 1 nobody nogroup   0 Mar 12 06:35 uevent
drwxr-xr-x 2 nobody nogroup   0 Mar 12 06:35 vulnerabilities
```

But then:
```
root@c1:~# umount /sys/devices/system/cpu/
root@c1:~# getconf _NPROCESSORS_ONLN
4
root@c1:~# getconf _NPROCESSORS_CONF
4
```

stgraber:
This is causing a fair bit of breakage in downstream software as those two sysconfs are commonly looked at by a variety of tools.

stgraber:
Broken getconf:
```
[...]

PR #525, opened 2022-03-13
More sysfs fixes

stgraber:


stgraber:
Note that as mentioned in #522, the underlying problem with `getdents64` returning `DT_UNKNOWN` as type for the various entries hasn't been resolved, the test for it was just incorrect.

This PR fixes the test which will now cause failures throughout, hopefully making the actual issue easier to track down and fix.

stgraber:
@brauner I think I found the reason for the failure:

https://github.com/lxc/lxcfs/blob/master/src/sysfs_fuse.c#L296

Basically the way things work is:
 - Generate entries for `cpuX` based on what's part of the cpuset mask
 - Then add everything else

The issue here is that the part generating the `cpuX` entries doesn't have access to the `dirent` struct as it's happening prior to iterating the actual directory. This means, we're calling `dir_filler` to create a generic entry, that entry isn't passed a `stat` struct and therefore doesn't have a `dt_type` set.

I think the right way to fix this is to:
 - Move the entire logic into the `readdir` loop, then when we hit a directory starting with `cpu` we check if it's part of the mask, if it's not, we skip it.
 - The logic you added for FUSE2 will then pass the correct `stat` struct through to FUSE which will correctly populate `dt_type` on FUSE2
 - We then need something similar for FUSE3 as the `dirent_filler/dirent_fillerat` code for FUSE3 doesn't have the logic needed to pass through a suitable `stat` struct

brauner:
> @brauner I think I found the reason for the failure:
> 
> https://github.com/lxc/lxcfs/blob/master/src/sysfs_fuse.c#L296
> 
> Basically the way things work is:
> 
> * Generate entries for `cpuX` based on what's part of the cpuset mask
> * Then add everything else
> 
> The issue here is that the part generating the `cpuX` entries doesn't have access to the `dirent` struct as it's happening prior to iterating the actual directory. This means, we're calling `dir_filler` to create a generic entry, that entry isn't passed a `stat` struct and therefore doesn't have a `dt_type` set.
> 
> I think the right way to fix this is to:
> 
> * Move the entire logic into the `readdir` loop, then when we hit a directory starting with `cpu` we check if it's part of the mask, if it's not, we skip it.
> * The logic you added for FUSE2 will then pass the correct `stat` struct through to FUSE which will correctly populate `dt_type` on FUSE2
> * We then need something similar for FUSE3 as the `dirent_filler/dirent_fillerat` code for FUSE3 doesn't have the logic needed to pass through a suitable `stat` struct

We don't need that fix for FUSE3. FUSE3 will fill in that info for us. That's why we pass `FUSE_FILL_DIR_PLUS`.

PR #526, opened 2022-03-13
sysfs_fuse: fixes

brauner:
Signed-off-by: Christian Brauner (Microsoft) <christian.brauner@ubuntu.com>


brauner:
I can't fully fix this for fuse3 versions that do not container the fix in https://github.com/libfuse/libfuse/pull/591

PR #527, opened 2022-03-13
init/meson: Use libdir instead of hardcoded /lib path

Foxboron:
Hardcoding `/lib` makes meson create a directory which would conflict on
distros with usrmerge as `/lib` is a symlink. We define `libdir` in the
top-level so we should be using that instead.

Signed-off-by: Morten Linderud <morten@linderud.pw>


brauner:
Failures are currently expected. :)

PR #539, opened 2022-05-21
sysfs: correct file size of /sys/devices/system/cpu/online.

Hunter1016:
Calculate accurate size of 'online' file in sys_getattr routine.

Signed-off-by: huteng.ht <huteng.ht@bytedance.com>


brauner:
Thank you!

PR #545, opened 2022-07-05
re-initialize library after reload

Blub:
When introducing versioned options, we started using fuse's
"init" callback in order to tell the library to set
`can_use_sys_cpu` and `has_versioned_opts` accordingly.

However, we forgot to also do this on a reload. Fix this by
simply calling `lxcfs_fuse_init()` in `do_reload()` as well.

Additionaly: ignore lxcfs_fuse_init()'s return value.
We just "passed through" the private_data from fuse which is
set via the `fuse_main()` call.

It's better to not leave this up to the library anyway in
order to make it easier to be fuse version agnostic in the
future.

Without this, issuing a reload to lxcfs would cause
files in `/sys/devices/system/cpu/` to be visible via
`readdir`, but accessing them would fail:

    ~ # ls /sys/devices/system/cpu/
    ls: /sys/devices/system/cpu/cpuidle: No such file or directory
    ls: /sys/devices/system/cpu/uevent: No such file or directory
    (...)

    ~ # echo /sys/devices/system/cpu/*
    /sys/devices/system/cpu/cpu0 /sys/devices/system/cpu/cpu1 (...)

    ~ # strace stat /sys/devices/system/cpu/cpu0
    lstat("/sys/devices/system/cpu/cpu0", 0x7ffdb2c57a00) = -1 ENOENT (No such file or directory)

### DIFF
--- a/src/lxcfs.c
+++ b/src/lxcfs.c
@@ -105,6 +105,22 @@ static int stop_loadavg(void)
 
 static volatile sig_atomic_t need_reload;
 
+static int lxcfs_init_library(void)
+{
+	char *error;
+	void *(*__lxcfs_fuse_init)(struct fuse_conn_info * conn, void * cfg);
+
+	dlerror();
+	__lxcfs_fuse_init = (void *(*)(struct fuse_conn_info * conn, void * cfg))dlsym(dlopen_handle, "lxcfs_fuse_init");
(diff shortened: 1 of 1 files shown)


brauner (approved):
Thank you! Excellent catch!

PR #546, opened 2022-07-05
replace opathdir with opendir_flags

Blub:
`opathdir` was used to replace `opendir` in order to ensure
`O_NOFOLLOW` and `O_CLOEXEC` were set, however it also added
`O_PATH` which prevents `readdir`/`getdents` to be used on
it, causing the `/sys/devices/system/cpu/<subdir>`
directories to be empty.

Instead, let's have an `opendir_flags` utility which simply
passed additional flags to the `open(..., O_DIRECTORY)` call
preceding `fdopendir()`.


brauner:
Thank you!

YuiIchi:
In what version will this feature be released?

stgraber:
5.0.1

PR #552, opened 2022-08-23
set oom_score_adj of lxcfs process to -1000

Hunter1016:
Disable oom killing entirely to minimize the hassle comes from
lxcfs exiting unexpectedly, e.g. the mountpoint got lost.

Signed-off-by: Teng Hu <huteng.ht@bytedance.com>

### DIFF
--- a/config/init/systemd/lxcfs.service.in
+++ b/config/init/systemd/lxcfs.service.in
@@ -5,6 +5,7 @@ Before=lxc.service
 Documentation=man:lxcfs(1)
 
 [Service]
+OOMScoreAdjust=-1000
 ExecStart=/usr/bin/lxcfs {{LXCFSTARGETDIR}}
 KillMode=process
 Restart=on-failure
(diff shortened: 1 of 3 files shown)


brauner (changes requested):
I think that `LXCFS` itself should not alter it oom score. This should be added to the service file via `OOMScoreAdjust`.

Blub:
Indeed, this could just go in `config/init/systemd/lxcfs.service.in`, but I don't know if any of the other init systems have an idiomatic way to do this, or if we should just ignore them?

Hunter1016:
> Indeed, this could just go in `config/init/systemd/lxcfs.service.in`, but I don't know if any of the other init systems have an idiomatic way to do this, or if we should just ignore them?

For upstart, add 'oom score -1000' to config/init/upstart/lxcfs.conf works.
For sysvinit, just 'echo -1000 > /proc/`cat ${PIDFILE}`/oom_score_adj' after start-stop-daemon in config/init/sysvinit/lxcfs also works.
Will forced-push soon.

stgraber (approved):

PR #555, opened 2022-09-12
init: respect --prefix when installing systemd unit files

tych0:
Signed-off-by: Tycho Andersen <tycho@tycho.pizza>

### DIFF
--- a/config/init/meson.build
+++ b/config/init/meson.build
@@ -2,7 +2,7 @@
 
 if 'systemd' in init_script
     systemd = dependency('systemd')
-    systemd_system_unit_dir = systemd.get_pkgconfig_variable('systemdsystemunitdir')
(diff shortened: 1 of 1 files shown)


tych0:
Hmm. Actually this isn't quite what I want, but e.g. the sysconfdirs have `--prefix` in them, so maybe we want this? Dunno.

stgraber (approved):

thmo:
I think this is wrong. lxcfs-5.0.3 installs the service file to  `DESTDIR/usr/usr/lib/systemd/system/lxcfs.service`. Note the double `/usr` in there.

Can we revert that?

tych0:
Fine with me.

thmo:
@stgraber ?

PR #557, opened 2022-10-27
sysfs: don't mask cpus in /sys/devices/system/cpu

tych0:
The kernel does not mask the cpu%d dirs when they are offlined:

(root) /sys/devices/system/cpu # cat online
0-7
(root) /sys/devices/system/cpu # chcpu -d 4
CPU 4 disabled
(root) /sys/devices/system/cpu # cat online
0-3,5-7
(root) /sys/devices/system/cpu # cat offline
4
(root) /sys/devices/system/cpu # ls -al
total 0
drwxr-xr-x 16 root root    0 Oct 25 20:42 .
drwxr-xr-x 10 root root    0 Oct 25 20:42 ..
drwxr-xr-x  7 root root    0 Oct 25 20:42 cpu0
drwxr-xr-x  7 root root    0 Oct 25 20:42 cpu1
drwxr-xr-x  7 root root    0 Oct 25 20:42 cpu2
drwxr-xr-x  7 root root    0 Oct 25 20:42 cpu3
drwxr-xr-x  5 root root    0 Oct 25 20:42 cpu4
drwxr-xr-x  7 root root    0 Oct 25 20:42 cpu5
drwxr-xr-x  7 root root    0 Oct 25 20:42 cpu6
drwxr-xr-x  7 root root    0 Oct 25 20:42 cpu7
drwxr-xr-x  2 root root    0 Oct 25 20:43 cpufreq
drwxr-xr-x  2 root root    0 Oct 26 15:19 cpuidle
drwxr-xr-x  2 root root    0 Oct 26 15:19 hotplug
-r--r--r--  1 root root 4096 Oct 25 20:42 isolated
-r--r--r--  1 root root 4096 Oct 25 20:43 kernel_max
-r--r--r--  1 root root 4096 Oct 26 15:19 modalias
-r--r--r--  1 root root 4096 Oct 26 15:19 offline
-r--r--r--  1 root root 4096 Oct 25 20:42 online
-r--r--r--  1 root root 4096 Oct 25 20:43 possible
drwxr-xr-x  2 root root    0 Oct 26 15:19 power
-r--r--r--  1 root root 4096 Oct 25 20:43 present
drwxr-xr-x  2 root root    0 Oct 26 15:19 smt
-rw-r--r--  1 root root 4096 Oct 25 20:42 uevent
drwxr-xr-x  2 root root    0 Oct 26 15:19 vulnerabilities

let's not mask them in lxcfs either. In particular, we have observed this causing problems with some JVMs' implementation of Runtime.getRuntime().availableProcessors().

This is a bit of a strange patch: it seems masking this dir was always incorrect, so we could go back to just not offering it as an lxcfs endpoint, and having people use sysfs' implementation directly. But maybe people are expecting it now, so I've left it as a proxy. Perhaps a more appropriate patch is to just delete it entirely and add an API extension note?

Signed-off-by: Tycho Andersen <tycho@tycho.pizza>


brauner:
If that's right then we should try and delete it?

tych0:
Sure, I think I removed most of it, let me know if I missed anything. The commit that added it added a bunch of infra as well that seems probably useful to keep.

tych0:
Yeah, this is the kind of failure I was afraid of. `getconf _NPROCESSORS_CONF` does a readdir() of /sys/devices/system/cpu, so if we go back to only exporting online, then that will fail. But if we only export online, people should only bind mount online, so it should pass.

tych0:
Anyway, I dropped the test for now.

stgraber:
I think we need to keep /sys/devices/system/cpu around, if only for upgrade purposes and have it show the host values.

tych0:
Ok, hopefully I updated it correctly :)

stgraber:
@mihalicyn does that look good to you?

mihalicyn:
LGTM

mihalicyn (approved):

stgraber (approved):

PR #566, opened 2022-11-24
meson: allow not installing init script

paper42:
Sometimes it's useful to not install an init script, for example if the installed init system is less common (runit in my case). Without this change, the build system installs a service file that then has to be removed manually.


stgraber:
Wouldn't `none` be more descriptive?

paper42:
sorry, I just found I can use `-Dinit-script=[]`

Issue #553, opened 2022-08-11
arm64: /proc/cpuinfo doesn't honour personality inside LXD container


cjwatson:

stgraber:
Moving over to LXCFS. It may take us a little while before we have manpower to put on this (we'll have a new hire on it, just not sure about start date yet).

Until then, I'd recommend unmounting /proc/cpuinfo in such environments. It will have the downside of possibly over-reporting the number of CPU cores available to some tools, but that's likely less problematic than the incorrect CPU flags.

cjwatson:
@stgraber Thanks for the suggestion.  I've proposed https://code.launchpad.net/~cjwatson/launchpad-buildd/+git/launchpad-buildd/+merge/428923 for that.

mihalicyn:
We can try to detect called process pid on the fuse daemon side, because we have pid in `struct fuse_in_header` structure. And then use it to obtain personality of the caller.

lanmarc77:
> @cjwatson can you try `umount /proc/cpuinfo` in the container?

I might have something related. On Raspbian I have a similar issue after switching to the 64bit Kernel. All containers are still 32bit. After the change multiple entries in /proc inside the containers were not updated. I also tested a 64bit container with the same result. What the entries had in common was their size of 4096bytes and not the usual 0byte. I implemented the following one liner to the startup process of every container which is a workaround for me:
`/usr/bin/find /proc/ -maxdepth 1 -size 4096c -exec /bin/umount {} \;`
I think it might be related to cpuinfo and is not limited to only this entry. But then the Raspbian Kernel is a bit special anyways. I hope someone finds this workaround useful.

mihalicyn:
@lanmarc77 this issue was fixed already in https://github.com/lxc/lxcfs/pull/567
You just need to update the lxcfs on your machines.

PR #567, opened 2022-11-28
cpuinfo with personality

mihalicyn:
An attempt to address issue https://github.com/lxc/lxcfs/issues/553

### DIFF
--- a/src/proc_fuse.c
+++ b/src/proc_fuse.c
@@ -24,6 +24,7 @@
 #include <sys/mman.h>
 #include <sys/mount.h>
 #include <sys/param.h>
+#include <sys/personality.h>
 #include <sys/socket.h>
 #include <sys/syscall.h>
 #include <sys/sysinfo.h>
@@ -1500,6 +1501,46 @@ static int proc_slabinfo_read(char *buf, size_t size, off_t offset,
 	return total_len;
 }
 
+static int proc_read_with_personality(int (*do_proc_read)(char *, size_t, off_t,
+			     struct fuse_file_info *), char *buf, size_t size, off_t offset,
+			     struct fuse_file_info *fi)
+{
+	struct fuse_context *fc = fuse_get_context();
+	__u32 host_personality = liblxcfs_personality(), caller_personality;
+	bool change_personality;
+	int ret, read_ret;
+
+	if (get_task_personality(fc->pid, &caller_personality) < 0)
+		return log_error(0, "Failed to get caller process (pid: %d) personality", fc->pid);
+
+	/* do we need to change thread personality? */
+	change_personality = host_personality != caller_personality;
+
+	if (change_personality) {
+		ret = personality(caller_personality);
+		if (ret == -1)
+			return log_error(0, "Call to personality(%d) failed: %s\n",
+				caller_personality, strerror(errno));
+
+		lxcfs_debug("task (tid: %d) personality was changed %d -> %d\n",
+				(int)syscall(SYS_gettid), ret, caller_personality);
+	}
+
+	read_ret = do_proc_read(buf, size, offset, fi);
+
+	if (change_personality) {
+		ret = personality(host_personality);
+		if (ret == -1)
+			return log_error(0, "Call to personality(%d) failed: %s\n",
+				host_personality, strerror(errno));
+
+		lxcfs_debug("task (tid: %d) personality was restored %d -> %d\n",
+				(int)syscall(SYS_gettid), ret, host_personality);
+	}
+
+	return read_ret;
+}
+
 __lxcfs_fuse_ops int proc_read(const char *path, char *buf, size_t size,
 			       off_t offset, struct fuse_file_info *fi)
 {
@@ -1514,7 +1555,7 @@ __lxcfs_fuse_ops int proc_read(const char *path, char *buf, size_t size,
 						  buf, size, offset, f);
 	case LXC_TYPE_PROC_CPUINFO:
 		if (liblxcfs_functional())
-			return proc_cpuinfo_read(buf, size, offset, fi);
+			return proc_read_with_personality(&proc_cpuinfo_read, buf, size, offset, fi);
 
 		return read_file_fuse_with_offset(LXC_TYPE_PROC_CPUINFO_PATH,
 						  buf, size, offset, f);
--- a/src/utils.c
+++ b/src/utils.c
@@ -529,6 +529,29 @@ int safe_uint64(const char *numstr, uint64_t *converted, int base)
 	return 0;
 }
 
+int safe_uint32(const char *numstr, uint32_t *converted, int base)
+{
+	char *err = NULL;
+	unsigned long uli;
+
+	while (isspace(*numstr))
+		numstr++;
+
+	if (*numstr == '-')
+		return -EINVAL;
+
+	errno = 0;
+	uli = strtoul(numstr, &err, base);
+	if (errno == ERANGE && uli == UINT32_MAX)
+		return -ERANGE;
+
+	if (err == numstr || *err != '\0')
+		return -EINVAL;
+
+	*converted = (uint32_t)uli;
+	return 0;
+}
+
 static int char_left_gc(const char *buffer, size_t len)
 {
 	size_t i;
--- a/src/utils.c
+++ b/src/utils.c
@@ -665,3 +665,29 @@ DIR *opendir_flags(const char *path, int flags)
 
 	return dirp;
 }
+
+int get_task_personality(pid_t pid, __u32 *personality)
+{
+	__do_close int fd = -EBADF;
+	int ret = -1;
+	char path[STRLITERALLEN("/proc//personality") + INTTYPE_TO_STRLEN(pid_t) + 1];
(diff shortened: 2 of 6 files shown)


mihalicyn:
Hm, `gettid()` was replaced by `syscall(SYS_gettid)` to make build on Ubuntu 18.04 happy.

src/utils.c
```diff
@@ -642,3 +642,43 @@ DIR *opendir_flags(const char *path, int flags)
...
+	__do_close int fd = -EBADF;
+	int ret = -1;
+	char path[STRLITERALLEN("/proc//personality") + INTTYPE_TO_STRLEN(pid_t) + 1];
+	/* seq_printf(m, "%08x\n", task->personality); */
+	char buf[8 + 1];
+
+	snprintf(path, sizeof(path), "/proc/%d/personality", pid);
+	fd = open(path, O_RDONLY | O_CLOEXEC);
+	if (fd < 0)
+		goto err;
```
brauner, line 657: I think you can just `return -1` here directly.

src/utils.c
```diff
@@ -642,3 +642,43 @@ DIR *opendir_flags(const char *path, int flags)
...
+
+	ret = read(fd, buf, sizeof(buf) - 1);
```
brauner, line 659: Please use `read_nointr()` from `utils.{c,h}`.

src/utils.c
```diff
@@ -642,3 +642,43 @@ DIR *opendir_flags(const char *path, int flags)
...
+	if (ret >= 0) {
+		buf[ret] = '\0';
+		*personality = atoi(buf);
```
brauner, line 662: In LXC we have

```c
int lxc_safe_uint(const char *numstr, unsigned int *converted)
{
        char *err = NULL;
        unsigned long int uli;

        while (isspace(*numstr))
                numstr++;

        if (*numstr == '-')
                return -EINVAL;

        errno = 0;
        uli = strtoul(numstr, &err, 0);
        if (errno == ERANGE && uli == ULONG_MAX)
                return -ERANGE;

        if (err == numstr || *err != '\0')
                return -EINVAL;

        if (uli > UINT_MAX)
                return -ERANGE;

        *converted = (unsigned int)uli;
        return 0;
}
```

apparently I've never ported them to LXCFS.

src/proc_cpuview.c
```diff
@@ -937,6 +966,28 @@ int proc_cpuinfo_read(char *buf, size_t size, off_t offset,
 		return total_len;
 	}
 
+	if (get_host_personality(&host_personality) < 0)
```
brauner, line 969: I don't think we should call this here. Instead we should call this once when the shared module is loaded in the function marked as constructor in `bindings.c`.

So something like the following (UNTESTED):

```diff
diff --git a/src/bindings.c b/src/bindings.c
index 422a622..5ff6a76 100644
--- a/src/bindings.c
+++ b/src/bindings.c
@@ -45,9 +45,15 @@ static bool can_use_swap;
 static bool can_use_sys_cpu;
 static bool has_versioned_opts;
 static bool memory_is_cgroupv2;
+static __u32 host_personality;

 static volatile sig_atomic_t reload_successful;

+__u32 liblxcfs_personality(void)
+{
+       return host_personality;
+}
+
 bool liblxcfs_functional(void)
 {
        return reload_successful != 0;
@@ -919,6 +925,11 @@ static void __attribute__((constructor)) lxcfs_init(void)
                goto broken_upgrade;
        }

+       if (get_host_personality(&host_personality) < 0) {
+               lxcfs_info("Failed to retrieve host personality");
+               goto broken_upgrade;
+       }
+
        reload_successful = 1;
        return;

diff --git a/src/bindings.h b/src/bindings.h
index e35423a..1624784 100644
--- a/src/bindings.h
+++ b/src/bindings.h
@@ -107,6 +107,7 @@ extern bool liblxcfs_can_use_swap(void);
 extern bool liblxcfs_memory_is_cgroupv2(void);
 extern bool liblxcfs_can_use_sys_cpu(void);
 extern bool liblxcfs_has_versioned_opts(void);
+extern __u32 liblxcfs_personality(void);

 static inline bool lxcfs_has_opt(struct lxcfs_opts *opts, lxcfs_opt_t opt)
 {
```

src/proc_cpuview.c
```diff
@@ -937,6 +966,28 @@ int proc_cpuinfo_read(char *buf, size_t size, off_t offset,
...
+		if (ret == -1) {
+			return log_error(0, "Call to personality(%d) failed: %s\n",
+				caller_personality, strerror(errno));
+		}
+
+		lxcfs_debug("task (tid: %d) personality was changed %d -> %d\n",
+				(int)syscall(SYS_gettid), ret, caller_personality);
+
+		personality_changed = true;
+	}
```
brauner, line 989: So I'm not completely sure but I think this doesn't work. We can't change the personality past this function lest we confuse subsequent callers. The thread is a worker thread so doesn't go away. IOW, at the end of this function you need to restore the original personality. One possibility is to wrap `proc_cpuinfo_read()` so am like:

```diff
diff --git a/src/proc_fuse.c b/src/proc_fuse.c
index 5416d43..2e6bc09 100644
--- a/src/proc_fuse.c
+++ b/src/proc_fuse.c
@@ -1513,8 +1513,11 @@ __lxcfs_fuse_ops int proc_read(const char *path, char *buf, size_t size,
                return read_file_fuse_with_offset(LXC_TYPE_PROC_MEMINFO_PATH,
                                                  buf, size, offset, f);
        case LXC_TYPE_PROC_CPUINFO:
-               if (liblxcfs_functional())
+               if (liblxcfs_functional()) {
+                       // assume personality
                        return proc_cpuinfo_read(buf, size, offset, fi);
+                       // restore original personality
+               }

                return read_file_fuse_with_offset(LXC_TYPE_PROC_CPUINFO_PATH,
                                                  buf, size, offset, f);
```

src/proc_cpuview.c
```diff
@@ -920,6 +947,8 @@ int proc_cpuinfo_read(char *buf, size_t size, off_t offset,
 	bool use_view;
 	char *cache = d->buf;
 	size_t cache_size = d->buflen;
+	__u32 host_personality, caller_personality;
+	call_cleaner(restore_personality) bool personality_changed = false;
```
brauner, line 951: Ah, I missed that. I think wrapping `proc_cpuinfo_read()` might be cleaner though.

brauner (changes requested):

src/utils.c
mihalicyn: ported, fixed ;-)

src/proc_cpuview.c
mihalicyn: Thanks! Have done. :-)

src/proc_cpuview.c
mihalicyn: I've introduced the wrapper function `proc_read_with_personality` for such cases.

src/utils.c
```diff
@@ -642,3 +642,43 @@ DIR *opendir_flags(const char *path, int flags)
...
+
+int get_task_personality(pid_t pid, __u32 *personality)
+{
+	__do_close int fd = -EBADF;
+	int ret = -1;
+	char path[STRLITERALLEN("/proc//personality") + INTTYPE_TO_STRLEN(pid_t) + 1];
+	/* seq_printf(m, "%08x\n", task->personality); */
+	char buf[8 + 1];
+
+	snprintf(path, sizeof(path), "/proc/%d/personality", pid);
```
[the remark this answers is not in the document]
mihalicyn: Agreed totally. I've ported this macro and used it. I'll also walk through other places in lxcfs where we use `snprintf` without checking retvalue and apply this macro in a separate PR.

mihalicyn:
Not related to this PR directly, but thanks to Christian I've noticed that in `lxcfs_init` function we combining `lxc_info` and `lxc_exit` calls on error paths, like this:
```
	init_ns = preserve_ns(pid, "mnt");
	if (init_ns < 0) {
		lxcfs_info("Failed to preserve initial mount namespace");
		goto broken_upgrade;
	}

	/* This function calls unshare(CLONE_NEWNS) our initial mount namespace
	 * to privately mount lxcfs cgroups. */
	if (!cgfs_setup_controllers()) {
		log_exit("Failed to setup private cgroup mounts for lxcfs");
		goto broken_upgrade;
	}
```

It seems strange because `log_exit` implies `exit` call in it. So, we don't reach goto in this case. Looks like a bug and should be addressed in another PR.

src/proc_fuse.c
```diff
@@ -1500,6 +1501,46 @@ static int proc_slabinfo_read(char *buf, size_t size, off_t offset,
...
+				caller_personality, strerror(errno));
+
+		lxcfs_debug("task (tid: %d) personality was changed %d -> %d\n",
+				(int)syscall(SYS_gettid), ret, caller_personality);
+	}
+
+	read_ret = do_proc_read(buf, size, offset, fi);
+
+	if (change_personality) {
+		ret = personality(host_personality);
```
brauner, line 1532: We usually just check  `(ret < 0)` instead of equal -1. But it's not that important.

brauner (approved):

Issue #565, opened 2022-11-03
Overlap used memory/cpu/uptime etc inside containers


gribchenko:
* Distribution:
 Ubuntu 
 * Distribution version:
 20.04.5 LTS
 * The output of "lxc info" or if that fails:
 lxc info ct7839
Name: ct7839
Status: RUNNING
Type: container
Architecture: x86_64
PID: 2404105
Created: 2022/10/25 06:19 EEST
Last Used: 2022/11/03 18:12 EET

Resources:
  Processes: 39
  Disk usage:
    root: 5.40GiB
  CPU usage:
    CPU usage (in seconds): 2
  Memory usage:
    Memory (current): 184.11MiB
    Memory (peak): 186.48MiB
  Network usage:
    eth0:
      Type: broadcast
      State: UP
      Host interface: veth17b24774
      MAC address: 00:16:3e:4f:f2:29
      MTU: 1500
      Bytes received: 13.33kB
      Bytes sent: 39.75kB
      Packets received: 154
      Packets sent: 319
      IP addresses:
        inet:  *.*.*.248/32 (global)
    lo:
      Type: loopback
      State: UP
      MTU: 65536
      Bytes received: 0B
      Bytes sent: 0B
      Packets received: 0
      Packets sent: 0
      IP addresses:
        inet:  127.0.0.1/8 (local)
        inet6: ::1/128 (local)

   * Kernel version:
    5.4.0-131-generic lxc/lxd#147-Ubuntu
   * LXC version:
   5.7

   * LXD version:
   5.7

   * Storage backend in use:

# Issue description
When run top inside containers

seen flapping info from other containers

top - 18:13:19 up 9 days, 13:49,  0 users,  load average: 0.00, 0.00, 0.00
Tasks:  22 total,   1 running,  21 sleeping,   0 stopped,   0 zombie
%Cpu(s):  0.1 us,  0.0 sy,  0.0 ni, 99.9 id,  0.0 wa,  0.0 hi,  0.0 si,  0.0 st
MiB Mem :  16384.0 total,  15026.1 free,    659.4 used,    698.4 buff/cache
MiB Swap:      0.0 total,      0.0 free,      0.0 used.  15724.6 avail Mem

    PID USER      PR  NI    VIRT    RES    SHR S  %CPU  %MEM     TIME+ COMMAND
      1 root      20   0  151508  10100   7928 S   0.0   0.1   0:00.26 systemd
[...]

PR #568, opened 2022-12-02
[RFC] cpuview: fix possible use-after-free in find_proc_stat_node

mihalicyn:
An attempt to guess what happens in:
https://github.com/lxc/lxcfs/issues/565
https://discuss.linuxcontainers.org/t/number-of-cpus-reported-by-proc-stat-fluctuates-causing-issues/15780/14

Signed-off-by: Alexander Mikhalitsyn <aleksandr.mikhalitsyn@canonical.com>


src/proc_cpuview.c
```diff
@@ -232,6 +234,31 @@ static struct cg_proc_stat *prune_proc_stat_list(struct cg_proc_stat *node)
 		if (!cgroup_supports("cpu", node->cg, "cpu.shares")) {
 			struct cg_proc_stat *cur = node;
 
+			/*
+			 * We need to ensure that no one referenced this node,
+			 * because we are going to remove it from the list and free memory.
+			 *
+			 * If we can't grub lock then just keep this node for now.
```
simondeziel, line 241: `s/grub/grab the/`

src/proc_cpuview.c
```diff
@@ -232,6 +234,31 @@ static struct cg_proc_stat *prune_proc_stat_list(struct cg_proc_stat *node)
 		if (!cgroup_supports("cpu", node->cg, "cpu.shares")) {
 			struct cg_proc_stat *cur = node;
 
+			/*
+			 * We need to ensure that no one referenced this node,
+			 * because we are going to remove it from the list and free memory.
```
brauner, line 239: In the long run one would probably want a refcount pattern for this, i.e., have users do:

lock()
refcount_inc_not_zero()
unlock()

and have writers/deleters do

lock()
refcount_dec_and_test()
unlock()

But that would require a bigger redesign of this thing. Really what you'd want is something like a userspace rcu library which exists already but anyway, looks good enough as a fix.

brauner (approved):

src/proc_cpuview.c
mihalicyn: >But that would require a bigger redesign of this thing. Really what you'd want is something like a userspace rcu library which exists already but anyway, looks good enough as a fix.

Agreed totally. We can take a look on that in near future :-)

mihalicyn:
I've force pushed with the change of commit message. I added some explanation just in case.

mihalicyn:
Hmm, test checks are not starting for some reason. Let's try yet another force push :-)

brauner (approved):

Issue #573, opened 2022-12-21
lxcfs crash on lxd 5.9 rev 24164


zrav:
Due to https://discuss.linuxcontainers.org/t/number-of-cpus-reported-by-proc-stat-fluctuates-causing-issues/15780 we are running LXD 5.9 revision 24164. After running a few days lxcfs crashed:

```
Dec 21 05:33:41 kernel: show_signal_msg: 14 callbacks suppressed
Dec 21 05:33:41 kernel: lxcfs[3219179]: segfault at 0 ip 00007f8084afdf81 sp 00007f8084a2e780 error 6 in libc-2.31.so[7f8084a94000+178000]
Dec 21 05:33:41 kernel: Code: 00 00 4c 89 ef 4c 89 4c 24 08 e8 3a 68 00 00 48 89 e9 4c 89 e2 48 89 ee 48 8d 05 2a d2 15 00 4c 89 ef 48 89 84 24 e8 00 00 00 <c6> 45 00 00 e8 06 7e 00 00 89 d9 4c 89 fa 4c 89 f6 4c 89 ef e8 c6
```
This is an Ubuntu 22.04.1 running kernel 5.15.0-56-generic on an AMD Epyc 7702P (128 thread) system with 512GB RAM.
```
lxc info
config:
  core.https_address: '[::]:8443'
  core.trust_password: true
  images.auto_update_interval: "0"
api_extensions:
- storage_zfs_remove_snapshots
- container_host_shutdown_timeout
- container_stop_priority
- container_syscall_filtering
- auth_pki
- container_last_used_at
- etag
- patch
- usb_devices
- https_allowed_credentials
- image_compression_algorithm
- directory_manipulation
- container_cpu_time
- storage_zfs_use_refquota
- storage_lvm_mount_options
- network
- profile_usedby
- container_push
- container_exec_recording
- certificate_update
- container_exec_signal_handling
- gpu_devices
- container_image_properties
- migration_progress
- id_map
- network_firewall_filtering
- network_routes
- storage
- file_delete
- file_append
- network_dhcp_expiry
- storage_lvm_vg_rename
- storage_lvm_thinpool_rename
- network_vlan
- image_create_aliases
- container_stateless_copy
- container_only_migration
- storage_zfs_clone_copy
- unix_device_rename
- storage_lvm_use_thinpool
- storage_rsync_bwlimit
- network_vxlan_interface
- storage_btrfs_mount_options
- entity_description
- image_force_refresh
[...]

PR #575, opened 2023-01-10
Handle NULL in releasedir

deleriux:
Whilst libfuse seems to indicate this should not occur when `nullpath_ok` is switched off, this does not always seem to be the case. A straightforward workaround is to handle the NULL case of `path` in `lxcfs_releasedir` such that no crash can occur in snaps.

Example crash we caught a few days ago:
```
Program terminated with signal SIGSEGV, Segmentation fault.
#0  0x00007fa33020bbee in ?? () at ../sysdeps/x86_64/multiarch/strcpy-evex.S:933 from /lib/x86_64-linux-gnu/libc.so.6
933     ../sysdeps/x86_64/multiarch/strcpy-evex.S: No such file or directory.
lxc/lxd#1  0x0000564ac32ea508 in lxcfs_releasedir (path=0x0, fi=0x7fa30fffec80) at ../src/src/lxcfs.c:774
lxc/lxd#2  0x00007fa3302c62b7 in ?? () from /snap/lxd/current/lib/x86_64-linux-gnu/libfuse3.so.3
lxc/lxd#3  0x00007fa3302cd683 in ?? () from /snap/lxd/current/lib/x86_64-linux-gnu/libfuse3.so.3
lxc/lxd#4  0x00007fa3302cfc59 in ?? () from /snap/lxd/current/lib/x86_64-linux-gnu/libfuse3.so.3
lxc/lxd#5  0x00007fa3302cb407 in ?? () from /snap/lxd/current/lib/x86_64-linux-gnu/libfuse3.so.3
lxc/lxd#6  0x00007fa33029d609 in ?? ()
lxc/lxd#7  0x0000000000000000 in ?? ()
```

Signed-off-by: Matthew Ife <75210448+deleriux@users.noreply.github.com>

### DIFF
--- a/src/lxcfs.c
+++ b/src/lxcfs.c
@@ -771,6 +771,13 @@ static int lxcfs_releasedir(const char *path, struct fuse_file_info *fi)
 {
 	int ret;
 
+        if (path == NULL) {
+                up_users();
+                ret = do_cg_releasedir(path, fi);
+                down_users();
+                return ret;
+        }
+	
 	if (strcmp(path, "/") == 0)
 		return 0;
 
@@ -1032,6 +1039,7 @@ static void *lxcfs_init(struct fuse_conn_info *conn)
 
 #if HAVE_FUSE3
 	cfg->direct_io = 1;
+	cfg->nullpath_ok = 0;
 #endif
 
 	return fuse_get_context()->private_data;


src/lxcfs.c
```diff
@@ -771,6 +771,13 @@ static int lxcfs_releasedir(const char *path, struct fuse_file_info *fi)
 {
 	int ret;
 
+        if (path == NULL) {
+                up_users();
+                ret = do_cg_releasedir(path, fi);
```
mihalicyn, line 776: why we are calling `do_cg_releasedir` here? Why not `do_sys_releasedir`? Without `path` we can't determine what to call here. I think we have to print out some error to the LXCFS log and return `-EINVAL`.

src/lxcfs.c
```diff
@@ -1032,6 +1039,7 @@ static void *lxcfs_init(struct fuse_conn_info *conn)
 
 #if HAVE_FUSE3
 	cfg->direct_io = 1;
+	cfg->nullpath_ok = 0;
```
mihalicyn, line 1042: I've asked that already https://github.com/lxc/lxd-pkg-snap/issues/110#issuecomment-1377472107
Did you catch the case when this option has a non-zero value? (I mean by printf-ing it's value or using gdb memory inspection).

mihalicyn (changes requested):

src/lxcfs.c
mihalicyn: >I am (potentially unsafely) assuming that whatever user data exists in fi will always be the same struct being released here.

Yep, that's what I'm trying to say. Right now that makes no difference to call `do_cg_releasedir` or `do_sys_releasedir` but in the future, this may be changed. We can't rely on that.

>The problem with not doing anything at all is may end up just slowly leaking memory.

Yep, but a memory leak combined with the warning message in logs is better than potentially incorrect memory free ;-)

>Passing -EINVAL will result in unexpected behaviour in userspace applications.

I've checked how this retvalue is handled in libfuse... and... it's just ignored. :-)
https://github.com/libfuse/libfuse/blob/33736958b61d90d9db9b03441103035316f09abc/lib/fuse.c#L1982
Take a look on `fuse_fs_releasedir()` calls.

>Spitting a log output might keep the process running but its pretty unlikely with the information kept in the userspace structure will reveal what the original path was that was rmdir'd. You could extend the userspace structure to add tracking though.

Yep, but this warning will be a sign of a problem. And yes, extending the log message with some additional info, in this case, is a good idea.

deleriux: > I've checked how this retvalue is handled in libfuse... and... it's just ignored. :-)
Thats great! I agree in that case leak/EINVAL is nicer.

If I can get the time approved I shall try to work out what the condition is that results in this behaviour as it might reveal something more fundamental going on.

mihalicyn:
Just to make a link https://github.com/lxc/lxcfs/issues/573

src/lxcfs.c
mihalicyn: BTW, we can use `(struct file_info)->type` to detect what to call `do_cg_releasedir` or `do_sys_releasedir`... Then we don't need to know the `path` at all in the release* callbacks!

deleriux: With a bit of hacking, I managed to resolve the userspace component of the coredumps I have in the backtraces.
These are 3 different crash examples.

```
(gdb) print *(struct file_info *)fi->fh                                                                 
$2 = {controller = 0x7f0408105af0 "systemd", 
  cgroup = 0x7f04081075e0 "lxc.payload.infra-app-test-31/system.slice/snap.mount", file = 0x0 <foo>, 
  type = 0, buf = 0x0 <foo>, buflen = 0, size = 779116910, cached = 1853189997}
```

And

```
(gdb) print *(struct file_info *)fi->fh
$1 = {controller = 0x7f02f4106a50 "devices", cgroup = 0x7f02f4118740 "lxc.payload.infra-app-test-33/system.slice/sssd-pac.socket", file = 0x0 <foo>, type = 0, buf = 0x0 <foo>, buflen = 0, 
  size = 778989426, cached = 1987208563}
```

And

```
(gdb) print *(struct file_info *)fi->fh
$4 = {controller = 0x7f9f90000ca0 "systemd", cgroup = 0x7f9f9000d470 "lxc.payload.infra-app-test-31/system.slice/system-getty.slice", file = 0x0 <foo>, type = 0, buf = 0x0 <foo>, buflen = 0, 
  size = 779318626, cached = 1987208563}
```

We use a lot of cgroup v1 namespaces on these hosts, however (which I have no backtraces for unfortunately) we have another that exhibits the same behaviour that is using cgroupv2 only).

In any case they are always cgroup files that crash in my case.

We pretty much can hammer our hosts (having 150+ active containers at at time).

I will update the PR with code tomorrow that handles all 3 cases cleanly.

mihalicyn:
Let's close this PR cause we have https://github.com/lxc/lxcfs/pull/577

PR #577, opened 2023-01-11
lxcfs: handle NULL path in lxcfs_releasedir/lxcfs_release

deleriux:
This fixes SEGV in `lxcfs_release/releasedir` when path=NULL. 
Avoid the use of paths doing the release and prefer to inspect the type `file_info` struct.
Print an error in the case no valid type can be detected.

### DIFF
--- a/src/bindings.h
+++ b/src/bindings.h
@@ -30,6 +30,8 @@
 /* Maximum number for 64 bit integer is a string with 21 digits: 2^64 - 1 = 21 */
 #define LXCFS_NUMSTRLEN64 21
 
+/* The definitions here are well-ordered. New values should go directly
+ * above LXC_TYPE_MAX only. */
 enum lxcfs_virt_t {
 	LXC_TYPE_CGDIR,
 	LXC_TYPE_CGFILE,
@@ -67,7 +69,14 @@ enum lxcfs_virt_t {
 
 	LXC_TYPE_SYS_DEVICES_SYSTEM_CPU_ONLINE,
 #define LXC_TYPE_SYS_DEVICES_SYSTEM_CPU_ONLINE_PATH "/sys/devices/system/cpu/online"
+	LXC_TYPE_MAX,
 };
+
+/* Macros below used to check the class from the file types above */
+#define LXCFS_TYPE_CGROUP(type) (type >= LXC_TYPE_CGDIR && type <= LXC_TYPE_CGFILE)
+#define LXCFS_TYPE_PROC(type) (type >= LXC_TYPE_PROC_MEMINFO && type <= LXC_TYPE_PROC_SLABINFO)
+#define LXCFS_TYPE_SYS(type) (type >= LXC_TYPE_SYS && type <= LXC_TYPE_SYS_DEVICES_SYSTEM_CPU_ONLINE)
+#define LXCFS_TYPE_OK(type) (type >= LXC_TYPE_CGDIR && type < LXC_TYPE_MAX)
 
 struct file_info {
 	char *controller;
--- a/src/lxcfs.c
+++ b/src/lxcfs.c
@@ -8,6 +8,7 @@
 #include <errno.h>
 #include <fcntl.h>
 #include <getopt.h>
+#include <inttypes.h>
 #include <libgen.h>
 #include <pthread.h>
 #include <sched.h>
@@ -63,6 +64,22 @@ static inline void users_lock(void)
 static inline void users_unlock(void)
 {
 	unlock_mutex(&user_count_mutex);
+}
+
+/* Returns file info type of custom type declaration carried
+ * in fuse_file_info */
+static inline enum lxcfs_virt_t file_info_type(struct fuse_file_info *fi)
+{
+	struct file_info *f;
+
+	f = INTTYPE_TO_PTR(fi->fh);
+	if (!f)
+		return -1;
+
+	if (!LXCFS_TYPE_OK(f->type))
+		return -1;
+
+	return f->type;
 }
 
 static pthread_t loadavg_pid = 0;
@@ -770,26 +787,33 @@ static int lxcfs_access(const char *path, int mode)
 static int lxcfs_releasedir(const char *path, struct fuse_file_info *fi)
 {
 	int ret;
+	enum lxcfs_virt_t type;
 
-	if (strcmp(path, "/") == 0)
-		return 0;
+	type = file_info_type(fi);
 
-	if (strncmp(path, "/cgroup", 7) == 0) {
+	if (LXCFS_TYPE_CGROUP(type)) {
 		up_users();
 		ret = do_cg_releasedir(path, fi);
 		down_users();
 		return ret;
 	}
 
-	if (strcmp(path, "/proc") == 0)
-		return 0;
-
-	if (strncmp(path, "/sys", 4) == 0) {
+	if (LXCFS_TYPE_SYS(type)) {
 		up_users();
 		ret = do_sys_releasedir(path, fi);
 		down_users();
 		return ret;
 	}
+
+	if (path) {
+		if (strcmp(path, "/") == 0)
+			return 0;
+		if (strcmp(path, "/proc") == 0)
+			return 0;
+	}
+
+	lxcfs_error("unknown file type: path=%s, type=%d, fi->fh=%" PRIu64,
+			path, type, fi->fh);
 
 	return -EINVAL;
 }
@@ -895,27 +919,33 @@ static int lxcfs_flush(const char *path, struct fuse_file_info *fi)
 static int lxcfs_release(const char *path, struct fuse_file_info *fi)
 {
 	int ret;
+	enum lxcfs_virt_t type;
 
-	if (strncmp(path, "/cgroup", 7) == 0) {
+	type = file_info_type(fi);
+
+	if (LXCFS_TYPE_CGROUP(type)) {
 		up_users();
 		ret = do_cg_release(path, fi);
 		down_users();
 		return ret;
 	}
 
-	if (strncmp(path, "/proc", 5) == 0) {
+	if (LXCFS_TYPE_PROC(type)) {
 		up_users();
 		ret = do_proc_release(path, fi);
 		down_users();
 		return ret;
 	}
 
-	if (strncmp(path, "/sys", 4) == 0) {
+	if (LXCFS_TYPE_SYS(type)) {
 		up_users();
 		ret = do_sys_release(path, fi);
 		down_users();
 		return ret;
 	}
+
+	lxcfs_error("unknown file type: path=%s, type=%d, fi->fh=%" PRIu64,
+			path, type, fi->fh);
 
 	return -EINVAL;
 }


deleriux:
This is the method I've approached this problem with we discussed yesterday.
I fleshed out the file type checks with some simple defines, then created a inline function to inspect the file type internally.

The release/releasedir code now primarily uses this mechanism to release files back to memory. It spits an error in the event it doesn't know what to do.

Flush is a noop so no work was required.

Might be a nice-to-have to extend the same checks into other callbacks as its a little quicker than a strcmp(), but that isn't the concern of this fix.

This still needs testing properly somewhere, I'll roll it out onto my own test systems tomorrow and check its not blowing up massively or leaking memory somehow.

src/lxcfs.c
```diff
@@ -1032,6 +1059,7 @@ static void *lxcfs_init(struct fuse_conn_info *conn)
 
 #if HAVE_FUSE3
 	cfg->direct_io = 1;
+	cfg->nullpath_ok = 0;
```
mihalicyn, line 1062: it makes no sense to add this here, as we discussed before, `cfg` is always initialized by zeroes. If you want to add this just to make the code more explicit, then we need to explicitly initialize all other options from `cfg` structure, IMHO.

src/lxcfs.c
```diff
@@ -895,28 +917,33 @@ static int lxcfs_flush(const char *path, struct fuse_file_info *fi)
...
 		return ret;
 	}
 
-	if (strncmp(path, "/sys", 4) == 0) {
+	if (LXCFS_TYPE_SYS(type)) {
 		up_users();
 		ret = do_sys_release(path, fi);
 		down_users();
-		return ret;
+		return ret;
```
mihalicyn, line 942: extra spaces after `;`

mihalicyn (changes requested):

mihalicyn:
This commit https://github.com/lxc/lxcfs/pull/577/commits/cb5f18af3328e31b88ed14f9cd080a43ea164018 looks strange because later you are removing files from it https://github.com/lxc/lxcfs/pull/577/commits/27f2cfd173be29bd8c7af648815c2f033ffceca2

deleriux:
Thanks. Whilst I'm fairly proficient in C my git-etiquette is unfortunately that great! I'll try to clean up my commits and remove the bits that are unnecessary.

As you can tell I took an indirect approach at first (to preserve the `dlopen()` consistency) but in the end I dont think it is necessary, extends the API without offering much to the end user and the extra runtime linker steps make it a slower constant approach than the `strcmp()` method.

I felt going to the effort of doing type-checks should at least potentially result in a slightly faster overall runtime experience.

src/bindings.h
```diff
@@ -31,6 +31,7 @@
 #define LXCFS_NUMSTRLEN64 21
 
 enum lxcfs_virt_t {
+	LXC_TYPE_START,
```
mihalicyn, line 34: I'm not sure that it's safe to change the start value of this enum. Cause we support live lxc library reload after this change `LXC_TYPE_CGDIR` becomes `1` instead of `0`. These values getting placed in the memory and kept between library reloads. So, let's not change this enum declaration at all. You can just place a comment inside enum definition which attracts developer attention to the macros `LXCFS_TYPE_CGROUP`, `LXCFS_TYPE_PROC`, `LXCFS_TYPE_SYS`.

src/bindings.h
```diff
@@ -67,8 +68,14 @@ enum lxcfs_virt_t {
 
 	LXC_TYPE_SYS_DEVICES_SYSTEM_CPU_ONLINE,
 #define LXC_TYPE_SYS_DEVICES_SYSTEM_CPU_ONLINE_PATH "/sys/devices/system/cpu/online"
+	LXC_TYPE_END,
 };
 
+#define LXCFS_TYPE_CGROUP(type) (type >= LXC_TYPE_CGDIR && type <= LXC_TYPE_CGFILE)
+#define LXCFS_TYPE_PROC(type) (type >= LXC_TYPE_PROC_MEMINFO && type <= LXC_TYPE_PROC_SLABINFO)
+#define LXCFS_TYPE_SYS(type) (type >= LXC_TYPE_SYS && type <= LXC_TYPE_SYS_DEVICES_SYSTEM_CPU_ONLINE)
+#define LXCFS_TYPE_OK(type) (type >= LXC_TYPE_START && type <= LXC_TYPE_END)
```
mihalicyn, line 77: `#define LXCFS_TYPE_OK(type) (type > LXC_TYPE_START && type < LXC_TYPE_END)`, correct?

mihalicyn (changes requested):

src/lxcfs.c
```diff
@@ -770,27 +786,33 @@ static int lxcfs_access(const char *path, int mode)
...
 		return ret;
 	}
 
+	if (LXCFS_TYPE_PROC(type)) 
+		return 0;
+
+	if (path && strcmp(path, "/") == 0) 
+		return 0;
+
+	lxcfs_error("lxcfs_releasedir() unknown file type: path=%s, type=%d\n",
```
mihalicyn, line 813: you don't need extra `\n` as `lxcfs_error` prints a new line automatically

src/lxcfs.c
```diff
@@ -895,28 +917,33 @@ static int lxcfs_flush(const char *path, struct fuse_file_info *fi)
...
-	if (strncmp(path, "/sys", 4) == 0) {
+	if (LXCFS_TYPE_SYS(type)) {
 		up_users();
 		ret = do_sys_release(path, fi);
 		down_users();
-		return ret;
+		return ret;	
 	}
 
+	lxcfs_error("lxcfs_release() unknown file type: path=%s, type=%d\n",
```
mihalicyn, line 945: `\n` is excess

mihalicyn (changes requested):

deleriux:
I've made the changes and cleaned up the commit history somewhat. I'll test the deployment out and see how it fares today.

src/bindings.h
deleriux: I saw and fixed this one earlier (the range checks being incorrect) might be best practice to keep the LXC_TYPE_END, it prevents needing to adjust LXCFS_TYPE_OK macro later on in the future and its pretty obvious to anyone going forwards not to add to the end of it.

src/lxcfs.c
deleriux: Fixed in latest commit.

src/lxcfs.c
deleriux: And again.

src/bindings.h
```diff
@@ -67,8 +67,15 @@ enum lxcfs_virt_t {
...
 	LXC_TYPE_SYS_DEVICES_SYSTEM_CPU_ONLINE,
 #define LXC_TYPE_SYS_DEVICES_SYSTEM_CPU_ONLINE_PATH "/sys/devices/system/cpu/online"
+	LXC_TYPE_END,
 };
 
+/* Macros below used to check the class from the file types above */
+#define LXCFS_TYPE_CGROUP(type) (type >= LXC_TYPE_CGDIR && type <= LXC_TYPE_CGFILE)
+#define LXCFS_TYPE_PROC(type) (type >= LXC_TYPE_PROC_MEMINFO && type <= LXC_TYPE_PROC_SLABINFO)
+#define LXCFS_TYPE_SYS(type) (type >= LXC_TYPE_SYS && type <= LXC_TYPE_SYS_DEVICES_SYSTEM_CPU_ONLINE)
+#define LXCFS_TYPE_OK(type) (type > LXC_TYPE_CGDIR && type < LXC_TYPE_END)
```
mihalicyn, line 77: `#define LXCFS_TYPE_OK(type) (type >= LXC_TYPE_CGDIR && type < LXC_TYPE_END)`?

src/bindings.h
deleriux: I agree here, totally didn't think of the case of reloading.

Generally we restart the service and rebind the mounts in the containers - there is a neat way to do this without restarting the containers, I'm still developing the software but am happy to share it later down the line.

src/lxcfs.c
deleriux: Fixed

src/lxcfs.c
deleriux: Dropped this change, agree its unnecessary to implies a cause where none exists.

src/lxcfs.c
```diff
@@ -895,28 +917,33 @@ static int lxcfs_flush(const char *path, struct fuse_file_info *fi)
...
 
-	if (strncmp(path, "/sys", 4) == 0) {
+	if (LXCFS_TYPE_SYS(type)) {
 		up_users();
 		ret = do_sys_release(path, fi);
 		down_users();
 		return ret;
 	}
 
+	lxcfs_error("lxcfs_release() unknown file type: path=%s, type=%d",
```
mihalicyn, line 945: `lxcfs_error` already prints the function name from where it's called ;)

src/lxcfs.c
```diff
@@ -770,27 +786,33 @@ static int lxcfs_access(const char *path, int mode)
...
 		return ret;
 	}
 
+	if (LXCFS_TYPE_PROC(type))
+		return 0;
+
+	if (path && strcmp(path, "/") == 0)
+		return 0;
+
+	lxcfs_error("lxcfs_releasedir() unknown file type: path=%s, type=%d",
```
mihalicyn, line 813: function name is already printed inside the `lxcfs_error`

mihalicyn (changes requested):

src/bindings.h
mihalicyn: agree with you :)

src/bindings.h
mihalicyn: > I agree here, totally didn't think of the case of reloading.
> 
> Generally we restart the service and rebind the mounts in the containers - there is a neat way to do this without restarting the containers, I'm still developing the software but am happy to share it later down the line.

great! I'm working on a way to recover the lxcfs from crashes another way. With preserving fuse connection between userspace and kernel. It's also not ready yet :)

mihalicyn:
Thanks @deleriux for working on that. I think it's also important to mention in the commit message and PR description that you are trying to fix SEGV with this patch :) Because right now, at first glance, it looks just like an improvement. But in fact, this is a bugfix.

mihalicyn:
cc @tych0

src/lxcfs.c
deleriux: Fixed

src/lxcfs.c
deleriux: Fixed

mihalicyn (commented):
the first line in the commit message should be like `lxcfs: handle NULL path in lxcfs_releasedir/lxcfs_release`.

And please rename PR accordingly.

deleriux:
`../src/lxcfs.c: 813: lxcfs_releasedir: unknown file type: path=/proc, type=-1`
Looks like there is more work here to do to handle base paths in `/proc` which aren't typically designated with a type themselves.

deleriux:
Fixed the latest minor niggle. There is no need to check for this in `lxcfs_release` since the inode is always a directory.
I also changed the commit message as reflected.

src/lxcfs.c
```diff
@@ -770,27 +786,36 @@ static int lxcfs_access(const char *path, int mode)
...
-
-	if (strncmp(path, "/sys", 4) == 0) {
+	if (LXCFS_TYPE_SYS(type)) {
 		up_users();
 		ret = do_sys_releasedir(path, fi);
 		down_users();
 		return ret;
 	}
 
+	if (LXCFS_TYPE_PROC(type))
```
mihalicyn, line 807: you don't need this check at all then. Cause for `/proc` directory we don't fill `fi` with any info in `lxcfs_opendir`:

```
static int lxcfs_opendir(const char *path, struct fuse_file_info *fi)
{
	int ret;

	if (strcmp(path, "/") == 0)
		return 0; // don't touch "fi"

	if (strncmp(path, "/cgroup", 7) == 0) {
		up_users();
		ret = do_cg_opendir(path, fi); // << fills "fi" with something
		down_users();
		return ret;
	}

	if (strcmp(path, "/proc") == 0)
		return 0; // don't touch "fi"

	if (strncmp(path, "/sys", 4) == 0) {
		up_users();
		ret = do_sys_opendir(path, fi); // << fills "fi" with something
		down_users();
		return ret;
	}

	return -ENOENT;
}
```

mihalicyn (changes requested):

src/lxcfs.c
deleriux: Excellent point. I should try to extend my head-logic a bit further ;). Updated.

mihalicyn (approved):
LGTM

Thanks!

mihalicyn:
And yes, one small suggestion. We can replace `lxcfs_error("unknown file type: path=%s, type=%d", path, type);` to `lxcfs_error("unknown file type: path=%s, type=%d, fi->fh=%" PRIu64, path, type, fi->fh);`

deleriux:
Fixed to include the pointer address. I assume you really only care if its a 0 or close to it though...

mihalicyn:
> Fixed to include the pointer address. I assume you really only care if its a 0 or close to it though...

in fact yes. But if we will have a coredump in a particular case this address may help to debug.

src/lxcfs.c
```diff
@@ -895,28 +919,34 @@ static int lxcfs_flush(const char *path, struct fuse_file_info *fi)
...
 
-	if (strncmp(path, "/sys", 4) == 0) {
+	if (LXCFS_TYPE_SYS(type)) {
 		up_users();
 		ret = do_sys_release(path, fi);
 		down_users();
 		return ret;
 	}
 
+	lxcfs_error("unknown file type: path=%s, type=%d, type=%" PRIu64,
```
mihalicyn, line 947: `type=` to `fi->fh=`

mihalicyn (changes requested):

src/lxcfs.c
deleriux: oof.. thanks.

mihalicyn (dismissed):

src/bindings.h
```diff
@@ -67,8 +67,15 @@ enum lxcfs_virt_t {
...
 	LXC_TYPE_SYS_DEVICES_SYSTEM_CPU_ONLINE,
 #define LXC_TYPE_SYS_DEVICES_SYSTEM_CPU_ONLINE_PATH "/sys/devices/system/cpu/online"
+	LXC_TYPE_END,
 };
 
+/* Macros below used to check the class from the file types above */
+#define LXCFS_TYPE_CGROUP(type) (type >= LXC_TYPE_CGDIR && type <= LXC_TYPE_CGFILE)
+#define LXCFS_TYPE_PROC(type) (type >= LXC_TYPE_PROC_MEMINFO && type <= LXC_TYPE_PROC_SLABINFO)
+#define LXCFS_TYPE_SYS(type) (type >= LXC_TYPE_SYS && type <= LXC_TYPE_SYS_DEVICES_SYSTEM_CPU_ONLINE)
+#define LXCFS_TYPE_OK(type) (type >= LXC_TYPE_CGDIR && type < LXC_TYPE_END)
```
brauner, line 77: This is error prone as it depends on ordering in the enum. The ordering isn't enforced however. So there needs to be at least a comment on top of the enum explaining this. Also, we should explicitly number the enum. This also makes it harder to accidently break ordering requirements.

brauner (changes requested):

src/bindings.h
mihalicyn: Technically, right now it's also not correct to add new values in between the enum. Cause we can perform reloading of `liblxcfs` in between `open` and `close` of the file descriptor. Then because enum values shift, some condition checks may become incorrect. For example https://github.com/lxc/lxcfs/blob/master/src/sysfs_fuse.c#L372

tych0:
Can you talk about what case the path would be null? At first glance this seems like a weird call for libfuse to make.

mihalicyn:
> Can you talk about what case the path would be null? At first glance this seems like a weird call for libfuse to make.

https://github.com/lxc/lxcfs/pull/575#discussion_r1066035718

tych0:
So I guess it's really the kernel, not libfuse, that's triggering these, given that they're just passthroughs from libfuse? I still don't understand why, but maybe it's a mystery for the ages...

mihalicyn:
> So I guess it's really the kernel, not libfuse, that's triggering these, given that they're just passthroughs from libfuse? I still don't understand why, but maybe it's a mystery for the ages...

As far as I know, the kernel doesn't know anything about the path, or, speaking more correctly, the kernel gives the path from `open*()` syscalls to fuse daemon only for the first time through a `FUSE_LOOKUP` request, then userspace should provide the kernel with inode number for this file descriptor. The later kernel will only send this number as file identification. So, libfuse handles this path resolve for our convenience. I feel that we have some bugs in libfuse, we have to add some workaround in our code to prevent crashes at least, IMHO.

tych0:
Yeah, that's what I was wondering about. Makes sense, thanks. Patch LGTM.

src/bindings.h
deleriux: I also did give this some thought beforehand. I thought some ways to do this could be..
1. re-order the enum numbering explicitly granting space for future files. IE
```
enum lxcfs_virt_t { 
  LXC_TYPE_CGROUP_BASE = 0x00000,
...
  LXC_TYPE_CGROUP_MAX,

  LXC_TYPE_PROC_BASE = 0x10000,
...
  LXC_TYPE_PROC_MAX,

  LXC_TYPE_SYS_BASE ,
...
  LXC_TYPE_SYS_MAX = 0x20000,
}
```
2. extend `file_info` to include a 'class' parameter, ditch the defines, remove `file_info_type` in favour of `file_info_class`.

For 1. Obviously you'd basically have to go through every file type already configured and put them in the right class, but at that point ordering no longer matters as you're relying on a different set of metadata to work out what goes where..

Also you could hijack the unused `fileinfo->buf` pointer that is and change it to a `unsigned char abi_version[sizeof(char *)]` that you could check going forwards. This would let you conditionally and smartly extend the runtime ABI yet be able to deal with previous versions of the ABI changing.
This is _of course_ assuming *buf is zeroed on initialization currently (version 0 implictly).
This _also_ felt like a whole load of architectural work for what is essentially a bugfix!

For 2. Its much simpler to implement and prevents in the future having to make a headache again.  That however breaks the ABI in a manner which would not survive a reload so didn't imagine that would get any traction..

deleriux: Just also thought you could actually do `LXC_TYPE_BASE = 0x00000`, copy all the file types as-is, then have new namespaces for each type and defines for which you use going forwards. That gives you both the means to maintain `reload` compatibility by checking if the type is in a 'base' namespace or the given correct namespace.

Its a little janky but probably simpler than doing some abi extension.

src/bindings.h
```diff
@@ -67,8 +67,15 @@ enum lxcfs_virt_t {
 
 	LXC_TYPE_SYS_DEVICES_SYSTEM_CPU_ONLINE,
 #define LXC_TYPE_SYS_DEVICES_SYSTEM_CPU_ONLINE_PATH "/sys/devices/system/cpu/online"
+	LXC_TYPE_END,
```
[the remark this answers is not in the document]
deleriux: Changed style from END to MAX.

src/bindings.h
brauner: For now just leaving a comment is fine. We just want to fix the immediate bug. Reworking it to something better should be a separate step. But feel free to follow this up with a series.

src/bindings.h
brauner: > I'm not sure that it's safe to change the start value of this enum. Cause we support live lxc library reload after this change `LXC_TYPE_CGDIR` becomes `1` instead of `0`. These values getting placed in the memory and kept between library reloads. So, let's not change this enum declaration at all.

I think we should explictly number it though. That's what I do nowadays for all enums I add becuase it eliminates a whole bunch of subtle bugs if you force the numbering.

src/bindings.h
deleriux: Added the requested comment above the ENUM indicating only appending to the bottom (above max).

src/bindings.h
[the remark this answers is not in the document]
deleriux: Added the requested comment also.

mihalicyn (approved):

brauner (approved):

PR #591, opened 2023-03-28
 cpuset cgroup path maybe different from cpu cgroup path in kubernetes

yowenter:
lxcfs use cpuset cgroup path to get cpu cgroup cfs_quota, cfs_period info, it works in most usecase.
however, if we deploy lxcfs on kubernetes with containerd runtime, the cpu and cpuset cgroup path is different.
the cpuset cgroup path  use prefix "/kubepods-pod"  while the cpu cgroup path use  prefix "/system.slice/containerd.service/kubepods-" .
in this case, the "/proc/cpuinfo" will be incorrect.

### DIFF
--- a/src/proc_cpuview.c
+++ b/src/proc_cpuview.c
@@ -496,20 +496,20 @@ static bool cfs_quota_disabled(const char *cg)
  * Return the maximum number of visible CPUs based on CPU quotas.
  * If there is no quota set, cpu number in cpuset value is returned.
  */
-int max_cpu_count(const char *cg)
+int max_cpu_count(const char *cpuset_cg, const char *cpu_cg)
 {
 	__do_free char *cpuset = NULL;
 	int rv, nprocs;
 	int64_t cfs_quota, cfs_period;
 	int nr_cpus_in_cpuset = 0;
 
-	if (!read_cpu_cfs_param(cg, "quota", &cfs_quota))
+	if (!read_cpu_cfs_param(cpu_cg, "quota", &cfs_quota))
 		cfs_quota = 0;
 
-	if (!read_cpu_cfs_param(cg, "period", &cfs_period))
+	if (!read_cpu_cfs_param(cpu_cg, "period", &cfs_period))
 		cfs_period = 0;
 
-	cpuset = get_cpuset(cg);
+	cpuset = get_cpuset(cpuset_cg);
 	if (cpuset)
 		nr_cpus_in_cpuset = cpu_number_in_cpuset(cpuset);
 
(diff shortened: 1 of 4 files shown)


yowenter:
@stgraber could you please take a review?

stgraber:
@mihalicyn

src/sysfs_fuse.c
```diff
@@ -57,7 +57,7 @@ static int do_cpuset_read(char *cg, char *buf, size_t buflen)
                 use_view = false;
 
         if (use_view)
-                max_cpus = max_cpu_count(cg);
+                max_cpus = max_cpu_count(cg, cg);
```
mihalicyn, line 60: Do you have any specific reason not to pass `cpuset` and `cpu` cgroup separately here too?

mihalicyn:
Hi @yowenter 

you need to signoff your patch https://github.com/lxc/lxcfs/blob/master/CONTRIBUTING.md#developer-certificate-of-origin

src/proc_cpuview.c
```diff
@@ -628,7 +628,7 @@ int cpuview_proc_stat(const char *cg, const char *cpuset,
 	}
 
 	/* Cannot use more CPUs than is available in cpuset. */
-	max_cpus = max_cpu_count(cg);
+	max_cpus = max_cpu_count(cg, cg);
```
mihalicyn, line 631: same

src/sysfs_fuse.c
yowenter: thanks for your review and advices.
At first I thought the func named `do_cpuset_read`, maybe it's not good to add `cpu_cg` argument,  it may kind of  break abstract layer. 

I've make some changes , please take a review again.

stgraber:
@mihalicyn can you review again?

src/proc_cpuview.c
```diff
@@ -972,8 +972,12 @@ int proc_cpuinfo_read(char *buf, size_t size, off_t offset,
 	cg = get_pid_cgroup(initpid, "cpuset");
 	if (!cg)
 		return read_file_fuse("proc/cpuinfo", buf, size, d);
-	prune_init_slice(cg);
```
mihalicyn, line 975: nit: IMHO, it's better to keep it there.

src/sysfs_fuse.c
```diff
@@ -108,8 +108,11 @@ static int sys_devices_system_cpu_online_read(char *buf, size_t size,
 	if (!cg)
 		return read_file_fuse("/sys/devices/system/cpu/online", buf, size, d);
 	prune_init_slice(cg);
+	cpu_cg = get_pid_cgroup(initpid, "cpu");
+	if (!cpu_cg)
+		return read_file_fuse("/sys/devices/system/cpu/online", buf, size, d);
```
mihalicyn, line 113: don't we need `prune_init_slice(cpu_cg);` here?

mihalicyn:
Hi @yowenter !

Please, squash commits into one ;-)

yowenter:
> Hi @yowenter !
> 
> Please, squash commits into one ;-)

@mihalicyn  hi, I've already fixed it, could u have a review again?

mihalicyn:
LGTM, except one small thing. Let's change commit title a little bit to be consistent with our style `cpuview: resolve cpu cgroup path separately from cpuset` (or like that).

mihalicyn:
@stgraber please, approve workflows to run

stgraber (approved):

PR #594, opened 2023-04-12
Partial support for FUSE_INTERRUPT

mihalicyn:
Let's start handling `FUSE_INTERRUPT` request in LXCFS.
This may help in a few scenarios:
- when buggy deadlock happens (user will be able to kill processes those are in fuse IO)
- when some fuse file becomes laggy (high lock contention?). We have seen big contention on some locks when kubernetes+monitoring software were used with LXCFS

I'm trying to be conservative and not make global changes at once. Let's just discuss about approach and convert one place at first. Then we will decide if it's helpful or not.

### DIFF
--- a/src/bindings.c
+++ b/src/bindings.c
@@ -149,6 +149,25 @@ static inline void store_unlock(void)
 	mutex_unlock(&pidns_store_mutex);
 }
 
+#define define_interruptible_lock(type, lockname, lockfn)           \
+        int lockname##_interruptible(type *l)                       \
+        {                                                           \
+                int ret = ETIMEDOUT;                                \
+                while (!fuse_interrupted() && (ret == ETIMEDOUT)) { \
+                        struct timespec deadline;                   \
+                        clock_gettime(CLOCK_REALTIME, &deadline);   \
+                        deadline.tv_sec += 1;                       \
+                        ret = lockfn(l, &deadline);                 \
+                }                                                   \
+                return -ret;                                        \
+        }
+
+define_interruptible_lock(pthread_mutex_t, mutex_lock, pthread_mutex_timedlock)
+define_interruptible_lock(pthread_rwlock_t, rwlock_rdlock, pthread_rwlock_timedrdlock)
+define_interruptible_lock(pthread_rwlock_t, rwlock_wrlock, pthread_rwlock_timedwrlock)
+
+#undef define_interruptible_lock
+
 /* /proc/       =    6
  *                +
  * <pid-as-str> =   INTTYPE_TO_STRLEN(pid_t)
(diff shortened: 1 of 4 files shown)


mihalicyn:
cc @tych0 @Blub

Blub:
Makes sense. Anything that can block should be cancellable. Though I do wonder why there has to actually be a signal involved if it's not really used? Or is it part of the kernel API?

mihalicyn:
Hi, Wolfgang!

>Though I do wonder why there has to actually be a signal involved if it's not really used? Or is it part of the kernel API?

technically it's a part of libfuse API https://github.com/libfuse/libfuse/commit/288ed4ebcea335c77793ee3d207c7466d55c4f71.

As far as I understand the idea this signal handler can be used on the filesystem side for some internal needs (for example to kill some thread that FS can create on top of threads those are created by libfuse). LXCFS is a quite simple in that so we just don't have any useful payload for this handler.

The most important part is that we have to set `SA_RESTART` flag otherwise we can break something accidentally. Because many syscalls are interrupted when signal arrives and return `-EINTR`. Userspace have to care about that by properly handling this errno. I can't be fully sure that we are in a good shape about that. That's why I think `SA_RESTART` is a good workaround, because then kernel will restart syscall (there are some exceptions but...) for us after `sigreturn`.

So, if some code is not ready to be interruptible then, with current approach, it will just ignore FUSE_INTERRUPT at all and continue to work as before.

mihalicyn:
> Looks reasonable enough to me. I wonder though - most of the files lxcfs touches are procfs or whatever; how often do lxcfs operations really get interrupted? I was thinking they'd be pretty quick, but maybe not?

yes, they are pretty quick. The problem happens if something is wrong with LXCFS code itself (deadlock, for example). Then end-users can't kill processes inside the container (even with SIGKILL!), because tasks those get stuck in reading fuse are in D-state.

tych0:
Ah ha, I see. Reminds me of: https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?h=5a8bee63b10f6f2f52f6d22e109a4a147409842a which I have seen a bit with lxcfs. I had just been aborting the whole filesystem in that case, so this does seem useful.

mihalicyn:
@stgraber don't we want to take this?

stgraber:
@mihalicyn I was skipping due to the RFC tag, if this is ready to go, clear it and I'll do another review and merge.

mihalicyn:
> @mihalicyn I was skipping due to the RFC tag, if this is ready to go, clear it and I'll do another review and merge.

ah, yep. Sorry about that. Fixed and ready for review ;-)

stgraber (approved):

PR #597, opened 2023-04-24
call fclose in get_procfile_size

choury:


Blub:
`f` is annotated with `__do_fclose`, so this should happen automatically

PR #601, opened 2023-07-07
support enable-host-diskstats option.

### LINKED ISSUE #599: /proc/diskstats information is inaccurate
## Background:
The `/proc/diskstats` information is assembled for a container by reading the blkio subsystem in the cgroup, in the `proc_diskstats_read` function.
## Problem:
However, the blkio cgroup subsystem can be inaccurate. For instance, if a container operates on the disk `/dev/sdc`, the host's `/proc/diskstats` may show a lot of operations on `/dev/sdc`. But the blkio cgroup may not present this information accurately, as `blkio.io_serviced_recursive`, `blkio.io_service_time_recursive`, and other data related to these operations may be empty.
## Proposed Solution:
To address this issue, one approach is to **read the host's `/proc/diskstats` to assemble the container's `/proc/diskstats`**. 
This can be accomplished by configuring independent disks for each container, such as using `/dev/sdc` as the data disk of container A and `/dev/sdd` as the data disk of container B. This practice is also widely adopted in the industry to isolate container resources.
To support this approach, lxcfs may need to isolate `/proc/partitions`(I don't know why the community doesn't support it so far) and modify the diskstats assembly logic to read the corresponding disk data from the host's `/proc/diskstats` according to the disk used by the container and reassemble it. 
**A better approach may be to use an option (`--enable-host-diskstats`) to switch between the old and new solutions, ensuring compatibility.**
Furthermore, given the limited accuracy of blkio cgroups in most cases, it may be redundant to use them to assemble /proc/diskstats.
Thank you for your attention！

peppaJoeng:
## new interface 
`--enable-host-diskstats`

## Experimental procedure
- Disk condition after partition：
```bash
$ lsblk
sdd                8:48   0  100G  0 disk 
├─sdd1             8:49   0   50G  0 part 
├─sdd2             8:50   0    1K  0 part 
└─sdd5             8:53   0   20G  0 part
```
- start lxcfs
```bash
$ lxcfs --enable-host-diskstats /var/lib/lxc/lxcfs/ &
```

- Start the container
```bash
$ docker run -tid -m 256m --device=/dev/sdd1:/dev/test -v /var/lib/lxc/lxcfs/proc/meminfo:/proc/meminfo:rw -v /var/lib/lxc/lxcfs/proc/diskstats:/proc/diskstats:rw --name test ubuntu bash
```
- View container's /proc/diskstats
```bash
$ docker exec -it test ls /dev/test
/dev/test
$ docker exec -it test cat /proc/diskstats
8       49 test 249 0 17928 24 75 1115 535153 152 0 169 180 51 0 104857608 3
$ cat /proc/diskstats 
   8       0 sda 23881 2843 2239558 980807 62612 124673 23769904 27578300 0 1989128 29020938 0 0 0 0 11198 461830
   8      32 sdc 1329 0 20890 1088 5376 71021 5013944 101385 0 52829 168482 21 0 41943048 1 277 66007
   8      48 sdd 1383 0 75670 143 104 1115 536836 643 0 813 1279 51 0 104857608 3 5 488
   8      49 sdd1 249 0 17928 24 75 1115 535153 152 0 169 180 51 0 104857608 3 0 0
   8      50 sdd2 10 0 68 0 0 0 0 0 0 10 0 0 0 0 0 0 0
   8      53 sdd5 312 0 20032 29 22 0 1667 2 0 82 32 0 0 0 0 0 0
```
- delete container
```
docker rm -f `docker ps -aq`
```
After enabling the enable-host-diskstats option, when the user accesses the /proc/diskstats file, the host /proc/diskstats data will be used instead of reading data from the cgroup. This option is suitable for scenarios where the container exclusively uses block/character devices.

fix #599

### DIFF
--- a/src/proc_fuse.c
+++ b/src/proc_fuse.c
@@ -28,6 +28,7 @@
 #include <sys/socket.h>
 #include <sys/syscall.h>
 #include <sys/sysinfo.h>
+#include <sys/sysmacros.h>
 #include <sys/vfs.h>
 
 #include "proc_fuse.h"
@@ -42,6 +43,25 @@
 #include "proc_loadavg.h"
 #include "proc_cpuview.h"
 #include "utils.h"
+
+static pthread_mutex_t container_dev_mutex = PTHREAD_MUTEX_INITIALIZER;
+static void lock_mutex(pthread_mutex_t *l)
+{
+	int ret;
+
+	ret = pthread_mutex_lock(l);
+	if (ret)
+		log_exit("%s - returned: %d\n", strerror(ret), ret);
+}
+
+static void unlock_mutex(pthread_mutex_t *l)
+{
+	int ret;
+
+	ret = pthread_mutex_unlock(l);
+	if (ret)
+		log_exit("%s - returned: %d\n", strerror(ret), ret);
+}
 
 struct memory_stat {
 	uint64_t hierarchical_memory_limit;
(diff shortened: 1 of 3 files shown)


mihalicyn:
Hi @peppaJoeng!

I'll review your PR on this week. Sorry about delay.

peppaJoeng:
Hi, @mihalicyn 
is there any progress？

src/proc_fuse.c
```diff
@@ -473,6 +493,147 @@ static void get_blkio_io_value(char *str, unsigned major, unsigned minor,
...
+		if (setns(nsfd, 0) < 0) {
+			lxcfs_error("Error setting mnt ns: %s\n", strerror(errno));
+			goto child_out;
+		}
+		dir = opendir("/dev");
+		if (dir == NULL) {
+			lxcfs_error("Error opening dir /dev: %s\n", strerror(errno));
+			goto child_out;
+		}
+		while ((ptr = readdir(dir)) != NULL) {
```
mihalicyn, line 566: here you are iterating over the `/dev` inside the container mount namespace to make a list of block devices which are used by the container, right?

What do you think about using `/proc/<pid>/mountinfo` file instead to get all used block devices? Speaking more precisely, this file allows you to get all block devices which are *mounted* inside the container mount namespace.

Another limitation here is that in the container there can be a few different mount namespaces.

peppaJoeng: Good idea! Thank you for your suggestion.

For the first question, I did this by going into the container's namespace and traversing the `/dev` directory underneath it.

I had thought about **using `/proc/PID/mountinfo` on the host side** (If I understand correctly) instead of going into the container namespace to achieve this functionality. But I have encountered some problems. 

1.  I cannot get the unmounted block device from this file; in fact, the block device mounted through `--device` cannot be obtained through the above method.
```bash
[root@localhost ~]# docker run -tid --name test --device /dev/sdb:/dev/xxx  rnd-dockerhub.huawei.com/official/ubuntu-arm64 bash
c82db9506ffaff30ce480852d3133cd4528a013d427372c77d91621edfb75809
[root@localhost ~]# docker inspect --format '{{.State.Pid}}'  c82db9506ffaff30ce480852d313
3325697
[root@localhost ~]# cat /proc/3325697/mountinfo | grep sdb
[root@localhost ~]# cat /proc/3325697/mountinfo | grep xxx
[root@localhost ~]# cat /proc/3325697/mountinfo | grep /dev
520 518 0:59 / /dev rw,nosuid - tmpfs tmpfs rw,seclabel,size=65536k,mode=755
521 520 0:60 / /dev/pts rw,nosuid,noexec,relatime - devpts devpts rw,seclabel,gid=5,mode=620,ptmxmode=666
530 523 0:37 /docker/c82db9506ffaff30ce480852d3133cd4528a013d427372c77d91621edfb75809 /sys/fs/cgroup/devices ro,nosuid,nodev,noexec,relatime master:11 - cgroup cgroup rw,seclabel,devices
540 520 0:57 / /dev/mqueue rw,nosuid,nodev,noexec,relatime - mqueue mqueue rw,seclabel
541 518 253:0 /var/lib/docker/containers/c82db9506ffaff30ce480852d3133cd4528a013d427372c77d91621edfb75809/resolv.conf /etc/resolv.conf rw,relatime - ext4 /dev/mapper/euleros-root rw,seclabel
542 518 253:0 /var/lib/docker/containers/c82db9506ffaff30ce480852d3133cd4528a013d427372c77d91621edfb75809/hostname /etc/hostname rw,relatime - ext4 /dev/mapper/euleros-root rw,seclabel
543 518 253:0 /var/lib/docker/containers/c82db9506ffaff30ce480852d3133cd4528a013d427372c77d91621edfb75809/hosts /etc/hosts rw,relatime - ext4 /dev/mapper/euleros-root rw,seclabel
544 520 0:56 / /dev/shm rw,nosuid,nodev,noexec,relatime - tmpfs shm rw,seclabel,size=65536k
464 520 0:60 /0 /dev/console rw,nosuid,noexec,relatime - devpts devpts rw,seclabel,gid=5,mode=620,ptmxmode=666
```

As you can see, I didn't get this device. However, I can get it in the following way:
```bash
[root@localhost ~]# docker exec -it test bash -c "ls /dev/xxx"
/dev/xxx
```

2. An important reason why I enter the `/dev` device to read the device name is that the device names in the container and the host are different.

Hope I understand you correctly. Welcome to exchange your ideas

botieking98:
Can anybody take some time to review this PR?

notlate:
Hi, @mihalicyn ，could you take a moment to review this code?

src/proc_fuse.c
mihalicyn: >I cannot get the unmounted block device from this file; in fact, the block device mounted through --device cannot be obtained through the above method.

yes, you can only get devices which are *mounted* inside the container using this method. `--device` option in Docker does *not* mount device, but just (I guess) performs a bind-mount of it inside the container.

mihalicyn (commented):
I like this feature and general, especially taking into account that it's a optional thing and if it's useful for some workloads I think we can take it after some rework and polishing.

Issue #608, opened 2023-08-26
Reading `/proc/cpuinfo` in a 32bit arm container on arm64 host fails for large numbers of CPUs


gibmat:
## Introduction

Earlier this year it was reported that on certain Debian armel/armhf builders, `/proc/cpuinfo` was empty ([bug report](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1036818)). Eventually this was tracked down to occurring in lxc containers running armel/armhf builds on arm64 hosts. (lxc version 5.0.2 and lxcfs version 5.0.3.)

Today, I finally had time to sit down, spin up VMs in QEMU and identify the root cause: the fix for #553 contains an edge case bug. When there are more than ~13 CPUs available to an arm64 host, any 32bit arm containers will fail to provide a populated `/proc/cpuinfo` due to exhausting the cache in `proc_cpuinfo_read()`.

## Problem

Code in the arm64 kernel ([link](https://github.com/torvalds/linux/blob/master/arch/arm64/kernel/cpuinfo.c#L191)) checks the process' personality when called, and if it's 32bit, introduces another line to each CPU reported in `/proc/cpuinfo` (for example `model name	: ARMv8 Processor rev 1 (v8l)`). That's about 40 bytes additional output per CPU that doesn't appear if called from a process with a 64bit personality. (The reported feature flags will also change, but at least in my QEMU setup the overall length of that line was about the same for 32bit and 64bit invocations.)

Currently, within `proc_cpuinfo_read()` the variable `cache_size` is set from the value of `d->buflen`, which is initialized in `src/proc_fuse.c` as `info->buflen = get_procfile_size(path) + BUF_RESERVE_SIZE`. `get_procfile_size()` simply gobbles up the specified file and reports its total size, but does so *without* accounting for any personality differences.

Therefore, when lxcfs sets up the expected size of the buffer for `/proc/cpuinfo` on an arm64 host, it unconditionally sets the buffer to expect the output of a 64bit system. When an armel/armhf container comes along, the result that is actually read from the host's `/proc/cpuinfo` is ~40 bytes longer per CPU which eats into `BUF_RESERVE_SIZE` (512 bytes), causing `proc_cpuinfo_read()` to fail if there are more than 512/~40 = ~13 CPUs on the host system.

## Reproducing

[...]

PR #609, opened 2023-09-04
proc: Fix /proc/cpuinfo not respecting personality

gibmat:
It was found that the personality within the container was not being properly respected, which for large numbers of CPUs would break reporting of `/proc/cpuinfo` in arm32 containers running on an arm64 host.

Tested on an arm64 QEMU VM; with this fix the 32bit container now properly reports `/proc/cpuinfo`. Shamelessly copied most of the code from the existing `proc_read_with_personality()`.

Fixes: #608.

### DIFF
--- a/src/proc_fuse.c
+++ b/src/proc_fuse.c
@@ -82,6 +82,45 @@ static off_t get_procfile_size(const char *path)
 	return answer;
 }
 
+static off_t get_procfile_size_with_personality(const char *path)
+{
+	struct fuse_context *fc = fuse_get_context();
+	__u32 host_personality = liblxcfs_personality(), caller_personality;
+	bool change_personality;
+	int ret;
+	off_t procfile_size_ret;
+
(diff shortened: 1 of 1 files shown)


stgraber:

mihalicyn:
> @mihalicyn can you take a look?

sure! My bad that I've skipped that one.

mihalicyn:
LGTM!

Big thanks for noticing, debugging and fixing, Mathias!

I'll do some refactoring in the following PR.

mihalicyn (approved):

gibmat:
Would it be possible to merge this in the next few days? I'd like to have the fix included in the Debian 12.2 point release (scheduled for Oct 7). Working back from that date, I need to have the updated package basically ready by the 23rd so there's plenty of time for the update to be approved and included by the release team.

stgraber (approved):

Issue #561, opened 2022-11-06
clang-15: incompatible integer to pointer conversion returning 'int' from a function with result type 'pthread_t'


RinCat:
Needs type conversion.

```
../lxcfs-5.0.2/src/proc_loadavg.c:606:10: error: incompatible integer to pointer conversion returning 'int' from a function with result type 'pthread_t' (aka 'struct __pthread *') [-Wint-conversion]
                return log_error(0, "Initialize hash_table fails in load_daemon!");
                       ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
../lxcfs-5.0.2/src/cgroups/../macro.h:58:2: note: expanded from macro 'log_error'
        ({                                          \
        ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
../lxcfs-5.0.2/src/proc_loadavg.c:611:10: error: incompatible integer to pointer conversion returning 'int' from a function with result type 'pthread_t' (aka 'struct __pthread *') [-Wint-conversion]
                return log_error(0, "Create pthread fails in load_daemon!");
                       ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
../lxcfs-5.0.2/src/cgroups/../macro.h:58:2: note: expanded from macro 'log_error'
        ({                                          \
        ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
1 warning and 2 errors generated.
```

mihalicyn:
@RinCat I can't reproduce this with clang 15/16. Are you using glibc or something else?

PR #610, opened 2023-09-05
proc_loadavg.c: Fix incompatible integer to pointer conversion

listout:
Newer compiler like Clang 16 and GCC 14 have certain error enabled by default, namely -Werror=incompatible-function-pointer-types. Which resutls in build error such as:
```
proc_loadavg.c:606:10: error: incompatible integer to pointer conversion returning int from a function with result type pthread_t
```

My patch supresses the error for now, but a proper fix would be better. Fist discovered on Gentoo linux (bug #894348).

Bug: https://bugs.gentoo.org/894348
Closes: https://github.com/lxc/lxcfs/issues/561

### DIFF
--- a/src/proc_loadavg.c
+++ b/src/proc_loadavg.c
@@ -636,12 +636,12 @@ pthread_t load_daemon(int load_use)
 
 	ret = init_load();
 	if (ret == -1)
-		return log_error(0, "Initialize hash_table fails in load_daemon!");
+		return (pthread_t)log_error(0, "Initialize hash_table fails in load_daemon!");
 
 	ret = pthread_create(&pid, NULL, load_begin, NULL);
 	if (ret != 0) {
 		load_free();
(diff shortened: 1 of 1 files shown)


stgraber:
@brauner @mihalicyn what do you think we should do to clean that up properly?

listout:
> Gentoo friends happy.

Hey! Just wanted to note that it's not just Gentoo, rather all new compiler (as I mentioned GCC14 and Clang 16). :)

mihalicyn:
> Hey! Just wanted to note that it's not just Gentoo, rather all new compiler (as I mentioned GCC14 and Clang 16). :)

haha, sure! But Gentoo of course the first in the line of distros in terms of having a new versions of compilers and stuff (and Fedora?) ;-)

mihalicyn (approved):

stgraber (approved):

PR #615, opened 2023-10-03
proc: fix MemAvailable in /proc/meminfo to exclude tmpfs files

kyeongy:
The "total_cache" from memory.stat of cgroup includes the memory used by tmpfs files ("total_shmem"). Considering it as available memory is wrong because files created on a tmpfs file system cannot be simply reclaimed.

So the available memory is calculated with the sum of:
 * Memory the kernel knows is free
 * Memory that contained in the kernel active file LRU, that can be reclaimed if necessary
 * Memory that is contained in the kernel non-active file LRU, that can be reclaimed if necessary


hallyn (approved):
Seems right to me.

mihalicyn:
>Considering it as available memory is wrong because files created on a tmpfs file system cannot be simply reclaimed.

It can be reclaimed in fact (to the swap).

But anyways, this change looks good to me.

Thanks!

mihalicyn (approved):

PR #621, opened 2023-12-01
Allow overriding RUNTIME_DIR with a cli flag.

sdab:
This adds a --runtime-dir flag which can override the RUNTIME_PATH (renamed RUNTIME_DEFAULT_PATH). This ended up being kind of tricky because of how lxcfslib can be reloaded and its use of a library constructor which mounts paths under the runtime path. In order read the cli flag and then set a variable in the library before it starts mounting things, I removed the constructor attribute and made it an explicit init call in order to override the runtime dir in between loading the library and initializing lxcfslib.

This allows me to run multiple instances of lxcfs without them sharing mounts/paths. Unfortunately, I think this a breaking lib change so it will require a restart.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -4,7 +4,7 @@
 project(
     'lxcfs',
     'c',
-    version: '5.0.0',
+    version: '6.0.0',
     license: 'LGPLv2+',
     default_options: [
         'b_colorout=always',


stgraber:
Commits don't appear to be correctly signed off.

sdab:
They are signed off with my personal email, but I guess it compares it to my git config: 
Expected "sdabdoub <sdabdoub@netflix.com>", but got "Sebastien Dabdoub <sebastien.dabdoub@gmail.com>".

Ill see if I can get them to match

src/bindings.c
```diff
@@ -973,5 +1003,21 @@ void *lxcfs_fuse_init(struct fuse_conn_info *conn, void *data)
...
+}
+
+bool set_runtime_path(const char* new_path)
+{
+        int pathlen;
+
+        if (new_path && strlen(new_path) < PATH_MAX) {
+                pathlen = strlen(new_path);
+                memcpy(runtime_path, new_path, pathlen);
+                runtime_path[pathlen] = '\0';
```
tych0, line 1016: Why not strcpy here?

stgraber:
@sdab not sure what's up with Github Actions on this one. Could you do a quick rebase or pointless amend of your last commit so you can force push and re-trigger the Github stuff?

sdab:
I did a blank amend + force push. It looks like it needs approval to run the tests/checks.

src/bindings.c
tych0: For copying the middle parts of strings like in your link it makes sense, but here all of this math can be replaced by a simple strcpy(), IIUC.

sdab:
Looks like my amend undid the sign-off? Should be good now

stgraber:
@mihalicyn can you review this please?

src/lxcfs.c
```diff
@@ -199,13 +232,19 @@ static void do_reload(bool reinit)
 
         dlopen_handle = dlopen(lxcfs_lib_path, RTLD_LAZY);
 	if (!dlopen_handle)
-		log_exit("%s - Failed to open liblxcfs.so", dlerror());
+		log_exit("%s - Failed to open liblxcfs.so at %s", dlerror(), lxcfs_lib_path);
 	else
 		lxcfs_debug("Opened %s", lxcfs_lib_path);
 
 good:
+        if (strlen(runtime_path) > 0 && !do_set_runtime_path(runtime_path)) {
```
mihalicyn, line 240: Do we really have to introduce two new exported symbols `set_runtime_path` and `lxcfslib_init` in there?

If you changing existing constructor to be a normal exported function you can also add an argument to it like this:
```
void lxcfslib_init(const char *new_path)
{
//... handle new_path here
```

As far as I understand you can't call `set_runtime_path` anywhere except *before* the `lxcfslib_init` anyways. Which is a good sign that this function should be a part of `lxcfslib_init`.

==

Second thing is that we are trying to maintain compatibility, which means that *old* LXCFS daemon should be able to load and use a new LXCFS library gracefully. With this change old LXCFS daemon will try to reload LXCFS library, constructor is missing, and it's fine and no error will be generated. But LXCFS will remain uninitialized and then it will crash or at least give fuse request failures to the user.
My idea is to move `lxcfs_init` (constructor) call to the `lxcfs_fuse_init(struct fuse_conn_info *conn, void *data)`, because it's always called almost at the same time as constructor should be called. In this case you don't need to add a new exported symbol `lxcfslib_init`. Just remove a `__attribute__((constructor))` from the `lxcfs_init` and then add an explicit call to the `lxcfs_fuse_init` for `lxcfs_init`.
Please, correct me if I'm wrong ;-)

mihalicyn: This solution is also not ideal because `lxcfs_fuse_init` is also relatively new (3 years, see https://github.com/lxc/lxcfs/commit/285aea407acd58811b3ecf53033497d0a25b83f9) exported symbol, which means that LXCFS daemons older than 3 years will still fail to properly load new LXCFS library. But it's better than nothing.

cc @brauner

src/bindings.c
sdab: moved to strcpy

src/lxcfs.c
```diff
@@ -1370,6 +1417,13 @@ int main(int argc, char *argv[])
 		goto out;
 	}
 
+        if (runtime_path_arg) {
+                runtime_path_len = strlen(runtime_path_arg);
+                memcpy(runtime_path, runtime_path_arg, runtime_path_len);
+                runtime_path[runtime_path_len] = '\0';
```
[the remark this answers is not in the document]
sdab: moved to strcpy

src/bindings.c
```diff
@@ -40,15 +41,30 @@
...
 static bool has_versioned_opts;
 static bool memory_is_cgroupv2;
 static __u32 host_personality;
+static char runtime_path[PATH_MAX] = DEFAULT_RUNTIME_PATH;
+
 
 static volatile sig_atomic_t reload_successful;
 
+
+static char* get_base_dir(void) {
```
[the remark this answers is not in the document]
sdab: done

src/lxcfs.c
```diff
@@ -1474,7 +1528,10 @@ int main(int argc, char *argv[])
 #endif
 
 	if (!pidfile) {
-		snprintf(pidfile_buf, sizeof(pidfile_buf), "%s/lxcfs.pid", RUNTIME_PATH);
+                if (runtime_path_arg)
+                        snprintf(pidfile_buf, sizeof(pidfile_buf), "%s/lxcfs.pid", runtime_path_arg);
```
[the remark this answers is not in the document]
sdab: done

src/lxcfs.c
sdab: I looked at this for a bit, but dont see a way around breaking backwards compatibility. Essentially the problem is that we need to send a new parameter (runtime_path) between the binary and the library.

I did however merge set_runtime_path into lxcfslib_init, so there is only one new symbol.

stgraber:
I'll let @mihalicyn provided a more detailed answer here but speaking of backward compat, a normal upgrade path on most distros will be that the library will be updated to 6.0 with a daemon running 5.0, that daemon will be asked to reload the library.

Doing so should not result in a crash as that would instantly break all containers :)

So we basically need the new lib to be fine with not being provided the extra data and behave as it currently would. But then if the extra data is provided, operate with the custom runtime dir.

Also, the new symbol you're introducing here is very specific to providing the runtime path, would it make sense to have it provide some kind of init struct so we don't need to add another symbol the next time we need something like this?

sdab:
ack, I wasnt sure how strict the backwards compatibility requirement was (I thought maybe on a major version, it was ok to break).

I have removed the runtime_path only export, unless you mean you want an more generic option struct as the function argument.

Ill have to think of a backwards compatible way to do this. A separate symbol might be helpful in that case and make it only called by a newer binary (is backwards compatibility only required for old binaries and not old libs? or is it both ways)

stgraber:
> Ill have to think of a backwards compatible way to do this. A separate symbol might be helpful in that case and make it only called by a newer binary (is backwards compatibility only required for old binaries and not old libs? or is it both ways)

We only care about the case where an old binary (already running) needs to talk to a new library, the other way around shouldn't be possible.

stgraber:
> I have removed the runtime_path only export, unless you mean you want an more generic option struct as the function argument.

Yeah, that. Having an init function which only takes a single string argument is a recipe for needing an initv2 function down the line when we need to pass something else, using a struct from the start saves us from having to keep adding new functions.

sdab:
Ok I reworked the code a bit to make it backwards compatible.

@stgraber I made use of the existing (and versioned) lxcfs_opts struct and bumped the version value to 2. Thus if an old binary loads this new lib then the version will be set to 1 and we will not attempt to read the new field. PTAL

sdab:
@mihalicyn let me know if I need to do anything else

stgraber:
@sdab looks like you have a small conflict and Github Actions failed to run the tests because of that. Can you rebase on main?

sdab:
Rebased

src/bindings.c
```diff
@@ -762,6 +762,11 @@ static int permute_prepare(void)
 		return -1;
 	}
 
+        if (!mkdir_p(new_runtime, 0755)) {
+		lxcfs_error("Failed to create dir %s\n", new_runtime);
+		return -1;
+        }
+
```
stgraber, line 769: What's going on with the indent here? Is it a mix of tabs/spaces?

src/bindings.h
```diff
@@ -111,6 +107,8 @@ struct lxcfs_opts {
 	bool swap_off;
 	bool use_pidfd;
 	bool use_cfs;
+        // As of opts version 2.
```
mihalicyn, line 110: we need to place all the new fields at the end of the struct

src/lxcfs.c
mihalicyn: I guess your current solution is correct and does not break compatibility against the recent LXCFS version.

mihalicyn: I guess your current solution is correct and does not break compatibility against the recent LXCFS version.

mihalicyn:
LGTM except small issue with a new field placement in the `struct lxcfs_opts`

Once this fixed I'll test this change locally and check that I can update LXCFS from the version in the `main` branch to the version in the `sdab:main` branch without crashes and issues.

src/bindings.c
sdab: Yeah, maybe something funky with my editor. Switched to tabs like (most) of the rest of the code.

src/bindings.h
sdab: Updated

sdab:
Addressed the comments, thanks @mihalicyn

mihalicyn:
LGTM, thanks for your work on this, Sebastien!

I'll make some small adjustments:
- squash 2 last commits into one
- small changes in the commit messages format to align with our current standards

mihalicyn (approved):

mihalicyn:
Updated version https://github.com/lxc/lxcfs/pull/645

mihalicyn:
> * instructions on how to run the tests and a backwards compatibility test

That's a good point. Today, I started to work on adding a new CI test to check forward-compatibility.
Hopefully, it will be ready for review tomorrow. ;-)

PR #624, opened 2023-12-14
typofix: fix incorrect printing in lxcfs help interface

peppaJoeng:
We occasionally found an error printing in the lxcfs help interface. In theory, -h was written incorrectly -n.

There is an exception error before modification
image

After modification：
image


stgraber (approved):
Great catch, thanks!

Issue #559, opened 2022-10-29
Mount hook fails on sysfs-less containers


allixx:
`share/lxcfs/lxc.mount.hook` fails to mount `/sys/devices/system/cpu` when target container doesn't have sysfs mounted, erroring the whole script due to `#/bin/sh -e` shebang, and preventing container start.

I propose making hook script more robust by adding a corresponding check.

https://github.com/allixx/lxcfs/commit/29ef6dce5fd5706d06a46f814ccfe04bde921abb

mihalicyn:
Fixed by @stgraber

PR #625, opened 2024-01-09
Prevent mount hook failure if /sys is not mounted.

amateur80lvl:
This patch fixes lxcfs failure when I don't want anything mounted in a container by placing 
```
lxc.mount.auto =
```
in the container's configuration file.

This should also fix https://github.com/lxc/lxcfs/issues/559


stgraber:
Your commit message appears to be missing the required `Signed-off-by` line.

amateur80lvl:
Signed-off-by does not make sense because: a) I'm an anonymous person; b) it's a trivial patch.

Issue #636, opened 2024-04-25
Errors related to personality retrieval with Yama


HorlogeSkynet:
Hello there :wave:

I've noticed literally thousands of errors related to personality retrieval, very likely due to Yama.

It looks like (lxcfs logs) :

```
lxcfs[XXXX]: ../src/proc_fuse.c: 94: get_procfile_size_with_personality: Failed to get caller process (pid: YYYY) personality
```

As Yama also reports unauthorized (prevented) accesses, it actually floods kernel logs as well :

```
kernel: ptrace attach of "REDACTED PROCESS NAME"[YYYY] was attempted by "/usr/bin/lxcfs /var/lib/lxcfs"[XXXX]
```

When we dig this a bit, we can read in `proc(5)` :

```
/proc/[pid]/personality (since Linux 2.6.28)
  This  read-only  file  exposes  the process's execution domain, as set by personality(2).
  The value is displayed in hexadecimal notation.

  Permission to access this file is governed by a ptrace access mode PTRACE_MODE_ATTACH_FSCREDS check; see ptrace(2)
```

I wanted to propose a patch but I'm not sure about the best approach here. Should we :

* revert #609 ?
* handle `EPERM` specifically so as to ignoring it ? (this wouldn't address Yama complaining about "ptrace attachment tentative")
* check whether Yama is enforced through `/proc/sys/kernel/yama/ptrace_scope` **before** even trying to get task personality (LBYL) ?
* something else ?

Thanks for your time, bye :pray: 

---

Setup : lxcfs 6.0.0 / Linux 6.8.4

stgraber:
How about setting up YAMA so lxcfs is allowed the action? 

All of the other options will cause incorrect cpuinfo data for some users so don't seem adequate.

This seems like an issue with the LSM configuration on the system rather than anything wrong with lxcfs.

stgraber:
Are you getting the issue above with the stock configuration of a major Linux distribution?

HorlogeSkynet:
The thing is Yama configuration is system-wide and not "per-process" (unless maybe we deal with `setcap` and capabilities, leveraging fs extended attributes).

Indeed I've got a hardened `ptrace.yama_scope` sysctl, which explains why I am one of the firsts to encounter this issue.

Do you think we should mention this somewhere in documentation ? Or as a regression in v6 changelog ?

Although I'd like to emphasize I think something should be handled in code, as :

[...]

PR #637, opened 2024-04-28
proc: checks system security policy before trying to get personalities

HorlogeSkynet:
---

See #636.

### DIFF
--- a/src/proc_fuse.c
+++ b/src/proc_fuse.c
@@ -145,8 +145,11 @@ __lxcfs_fuse_ops int proc_getattr(const char *path, struct stat *sb)
 	    strcmp(path, "/proc/swaps")		== 0 ||
 	    strcmp(path, "/proc/loadavg")	== 0 ||
 	    strcmp(path, "/proc/slabinfo")	== 0) {
-		if (liblxcfs_functional())
+		if (liblxcfs_functional()) {
+			if (!is_ptrace_allowed())
+				return log_error(-EACCES, RESTRICTED_YAMA_PTRACE_POLICY);
 			sb->st_size = get_procfile_size_with_personality(path);
+		}
 		else
 			sb->st_size = get_procfile_size(path);
 		sb->st_mode = S_IFREG | 00444;
@@ -206,8 +209,11 @@ __lxcfs_fuse_ops int proc_open(const char *path, struct fuse_file_info *fi)
 
 	info->type = type;
 
-	if (liblxcfs_functional())
+	if (liblxcfs_functional()) {
+		if (!is_ptrace_allowed())
+			return log_error(-EACCES, RESTRICTED_YAMA_PTRACE_POLICY);
 		info->buflen = get_procfile_size_with_personality(path) + BUF_RESERVE_SIZE;
+	}
 	else
 		info->buflen = get_procfile_size(path) + BUF_RESERVE_SIZE;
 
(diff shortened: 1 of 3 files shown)


mihalicyn:
Dear @HorlogeSkynet,

please add Signed-off-by tag to your commit message.

src/utils.h
```diff
@@ -25,6 +25,12 @@
 #define SEND_CREDS_NOTSK 1
 #define SEND_CREDS_FAIL 2
 
+#define SYS_FS_KERNEL_YAMA_PTRACE_SCOPE "/sys/kernel/yama/ptrace_scope"
+#define YAMA_PTRACE_SCOPE_CLASSIC 0
```
mihalicyn, line 29: Let's align these mode names with what we have in Linux kernel https://github.com/torvalds/linux/blob/98369dccd2f8e16bf4c6621053af7aa4821dcf8e/security/yama/yama_lsm.c#L23

src/utils.c
```diff
@@ -691,3 +691,42 @@ int get_task_personality(pid_t pid, __u32 *personality)
...
+	access mode check (see PROC(5)).
+*/
+bool is_ptrace_allowed(void)
+{
+	static int yama_ptrace_scope = -1;
+
+	__u32 buf_u32;
+	__do_close int fd = -EBADF;
+	int ret = -1;
+	char buf[8 + 1];
```
mihalicyn, line 707: `char buf[INTTYPE_TO_STRLEN(uint32_t)];`

src/utils.c
```diff
@@ -691,3 +691,42 @@ int get_task_personality(pid_t pid, __u32 *personality)
...
+		fd = open(SYS_FS_KERNEL_YAMA_PTRACE_SCOPE, O_RDONLY | O_CLOEXEC);
+		if (fd >= 0) {
+			ret = read_nointr(fd, buf, sizeof(buf) - 1);
+			if (ret >= 0) {
+				buf[ret] = '\0';
+				if (safe_uint32(buf, &buf_u32, 16) < 0)
+					return log_error(true, "Failed to read Yama ptrace scope %s", buf);
+
+				yama_ptrace_scope = buf_u32;
+			}
```
mihalicyn, line 723: Let's print something in the logs in case when `read_nointr` fails.

mihalicyn (commented):
LGTM, except a few small nits. Thanks!

src/proc_fuse.c
```diff
@@ -145,8 +145,11 @@ __lxcfs_fuse_ops int proc_getattr(const char *path, struct stat *sb)
 	    strcmp(path, "/proc/swaps")		== 0 ||
 	    strcmp(path, "/proc/loadavg")	== 0 ||
 	    strcmp(path, "/proc/slabinfo")	== 0) {
-		if (liblxcfs_functional())
+		if (liblxcfs_functional()) {
+			if (!is_ptrace_allowed())
+				return log_error(-EACCES, RETRICTED_YAMA_PTRACE_POLICY);
```
stgraber, line 150: Should probably be RESTRICTED rather than RETRICTED

stgraber: Though that likely doesn't matter too much as we'll want to use the much simpler approach that's not yama-specific anyways :)

HorlogeSkynet: Fixed, for posterity :upside_down_face:

PR #638, opened 2024-04-29
lxcfs: per instance configuration

mihalicyn:
This adds ability to set per-container configuration parameters (boolean).

This can be very useful in some workloads if system administrator (or container administrator) want's to have some specific LXCFS configuration in some instances, while having defaults on the other ones.

Change is forward compatible and LXCFS can be updated with live reload. Nothing will crash. Obviously, to make new configuration sub-tree accessible full daemon reload is required.


stgraber:
I did a first pass on this, basically looking at the various commits and what's being modified.
General change seems fine but I'll take a closer look later.

For now, a few things I think that should be done:
 - Split off all the bugfixes/refactor/cleanup changes and send those through a separate PR as we'll want to backport them to stable
 - Rebase that branch which should then basically just be the per-instance config stuff
 - Try to find a way to have this be tested by GH actions

hallyn:
Looks like some good stuff.  Will take some time to review, though.  Would be cool if there was an easy way on github to say "patch 1/N is approved and separate, merge it".  Perhaps conceptually the thing to do here (I'm not asking you to change it, just thinking...) would be to create a github project, have an issue for each independent change in this pr, and a pr for each issue.  I say this because if you *just* made this N PRs, i'd possibly be less likely to look at the earlier ones.

Anyway, thanks, I intend to take a close look.

PR #640, opened 2024-05-01
Fix lxcfs read null

### LINKED ISSUE #599: /proc/diskstats information is inaccurate

### LINKED ISSUE #635: lxcfs: handle NULL in lxcfs_read (segfault at 0, code=killed, status=11/SEGV)
os: `ubuntu 5.15.0-52`
lxcfs version: `4.0.11`

lxcfs is killed by `11/SEGV` signal, the `syslog` is as below:
```
Apr 18 22:54:43 10-169-58-129 kernel: [10829437.241075] lxcfs[197879]: segfault at 0 ip 00007f101184cf81 sp 00007f0fe9ffa790 error 6 in libc-2.31.so[7f10117e3000+178000]
Apr 18 22:54:43 10-169-58-129 systemd[1]: lxcfs.service: Main process exited, code=killed, status=11/SEGV
Apr 18 22:54:43 10-169-58-129 systemd[1]: var-lib-lxcfs.mount: Succeeded.
Apr 18 22:54:43 10-169-58-129 systemd[1]: lxcfs.service: Failed with result 'signal'.
Apr 18 22:54:43 10-169-58-129 systemd[1]: lxcfs.service: Consumed 2d 41min 37.091s CPU time.
Apr 18 22:54:43 10-169-58-129 systemd[1]: lxcfs.service: Scheduled restart job, restart counter is at 1.
Apr 18 22:54:43 10-169-58-129 systemd[1]: lxcfs.service: Consumed 2d 41min 37.091s CPU time.
```

the core dump explained by `gbd` from `/var/crash` folder is as below:
SeaTalk_IMG_20240424_171648

The code of `lxcfs.c:778` is here: https://github.com/lxc/lxcfs/blob/lxcfs-4.0.11/src/lxcfs.c#L778

A similar issue about 'NULL path in lxcfs_releasedir/lxcfs_release' fix: https://github.com/lxc/lxcfs/pull/577

DevonSchwartz:
We used the macros that were implemented for lxcfs_release() to fix the NULL path issue. This will replace the existing strcmp solution. 

These macros have already been merged into the main branch to resolve the lxcfs_release() issue. 

fix #635

### DIFF
--- a/src/lxcfs.c
+++ b/src/lxcfs.c
@@ -743,6 +743,9 @@ static int lxcfs_readdir(const char *path, void *buf, fuse_fill_dir_t filler,
 #endif
 {
 	int ret;
+	enum lxcfs_virt_t type;
+	
+	type = file_info_type(fi);
 
 	if (strcmp(path, "/") == 0) {
 		if (dir_filler(filler, buf, ".", 0) != 0 ||
@@ -755,21 +758,21 @@ static int lxcfs_readdir(const char *path, void *buf, fuse_fill_dir_t filler,
 		return 0;
 	}
 
-	if (cgroup_is_enabled && strncmp(path, "/cgroup", 7) == 0) {
+	if (cgroup_is_enabled && LXCFS_TYPE_CGROUP(type)) {
 		up_users();
 		ret = do_cg_readdir(path, buf, filler, offset, fi);
 		down_users();
 		return ret;
 	}
 
-	if (strcmp(path, "/proc") == 0) {
+	if (LXCFS_TYPE_PROC(type)) {
 		up_users();
 		ret = do_proc_readdir(path, buf, filler, offset, fi);
 		down_users();
 		return ret;
 	}
 
-	if (strncmp(path, "/sys", 4) == 0) {
+	if (LXCFS_TYPE_SYS(type)) {
 		up_users();
 		ret = do_sys_readdir(path, buf, filler, offset, fi);
 		down_users();
@@ -876,27 +879,33 @@ static int lxcfs_read(const char *path, char *buf, size_t size, off_t offset,
 		      struct fuse_file_info *fi)
 {
 	int ret;
+	enum lxcfs_virt_t type;
+	
+	type = file_info_type(fi);
 
-	if (cgroup_is_enabled && strncmp(path, "/cgroup", 7) == 0) {
+	if (cgroup_is_enabled && LXCFS_TYPE_CGROUP(type)) {
 		up_users();
 		ret = do_cg_read(path, buf, size, offset, fi);
 		down_users();
 		return ret;
 	}
 
-	if (strncmp(path, "/proc", 5) == 0) {
+	if (LXCFS_TYPE_PROC(type)) {
 		up_users();
 		ret = do_proc_read(path, buf, size, offset, fi);
 		down_users();
 		return ret;
 	}
 
-	if (strncmp(path, "/sys", 4) == 0) {
+	if (LXCFS_TYPE_SYS(type)) {
 		up_users();
 		ret = do_sys_read(path, buf, size, offset, fi);
 		down_users();
 		return ret;
 	}
+
+	lxcfs_error("unknown file type: path=%s, type=%d, fi->fh=%" PRIu64,
+		path, type, fi->fh); 
 
 	return -EINVAL;
 }
@@ -905,15 +914,18 @@ int lxcfs_write(const char *path, const char *buf, size_t size, off_t offset,
 		struct fuse_file_info *fi)
 {
 	int ret;
+	enum lxcfs_virt_t type;
+	
+	type = file_info_type(fi);
 
-	if (cgroup_is_enabled && strncmp(path, "/cgroup", 7) == 0) {
+	if (cgroup_is_enabled && LXCFS_TYPE_CGROUP(type)) {
 		up_users();
 		ret = do_cg_write(path, buf, size, offset, fi);
 		down_users();
 		return ret;
 	}
 
-	if (strncmp(path, "/sys", 4) == 0) {
+	if (LXCFS_TYPE_SYS(type)) {
 		up_users();
 		ret = do_sys_write(path, buf, size, offset, fi);
 		down_users();


stgraber:
@mihalicyn can you review this one please?

mihalicyn:
Hi @DevonSchwartz 

Thanks for your PR! Please, can you explain how this is connected with #599 ? Why have you mentioned this issue?

DevonSchwartz:
I made a mistake linking that issue. I meant to link this issue instead. I will go ahead and fix that. Sorry for the confusion.

mihalicyn:
Next question is why are you change only `lxcfs_read`, but no touching other functions.

Also, can you explain the reason of this changes:
https://github.com/lxc/lxcfs/pull/640/commits/7117464add690a8f7e3bb255158382aa4d47c996
here, you are changing `do_proc_read` -> `do_cg_read`
https://github.com/lxc/lxcfs/pull/640/commits/1b8d493c82a083c55d66e788aabe6dd8376ffe91
then, `do_cg_read` -> `do_proc_read`

DevonSchwartz:
I change lxcfs_read because that was the only function specified. I did not want to change other functions that were not in the scope of the issue
We changed those because we were accidentally calling do_cg_read() for every case of LXCFS_TYPE, instead of the correct function. For example, when LXCFS_TYPE_PROC(type) was true, we should not have called do_cg_read. This issue came up when we compiled locally.

mihalicyn:
Ah, so, this PR is an attempt to fix https://github.com/lxc/lxcfs/issues/635?

mihalicyn:
> We changed those because we were accidentally calling do_cg_read() for every case of LXCFS_TYPE, instead of the correct function. For example, when LXCFS_TYPE_PROC(type) was true, we should not have called do_cg_read. This issue came up when we compiled locally.

You need to squash all (3) commits into one then (https://docs.github.com/en/get-started/using-git/about-git-rebase). It *is* a good practice to have one commit per logical change. But in your case you have two commits (https://github.com/lxc/lxcfs/commit/7117464add690a8f7e3bb255158382aa4d47c996 and https://github.com/lxc/lxcfs/commit/1b8d493c82a083c55d66e788aabe6dd8376ffe91) which do the opposite thing (first - introduces a wrong change and second reverts it). This is not something we want. ;-)

mihalicyn:
In general, I can't see how this change can fix issue in #635, because in there we have a pretty-pretty weird stack. It looks like something is terribly wrong with the dynamic symbol resolution. It is not the same bug as we had with `lxcfs_release`.

DevonSchwartz:
Just squashed the commits.

src/lxcfs.c
```diff
@@ -876,28 +876,34 @@ static int lxcfs_read(const char *path, char *buf, size_t size, off_t offset,
 		      struct fuse_file_info *fi)
 {
 	int ret;
+	enum lxcfs_virt_t type;
+	
+	type = file_info_type(fi);
 
-	if (cgroup_is_enabled && strncmp(path, "/cgroup", 7) == 0) {
+	if (LXCFS_TYPE_CGROUP(type)) {
```
mihalicyn, line 883: `cgroup_is_enabled && LXCFS_TYPE_CGROUP(type)`

mihalicyn (changes requested):
small nit

mihalicyn:
Please, edit commit message:
```
 Combine last 3 commits into one commit that changes the macros

Changed the strcmp in lxcfs_read() to macros

Signed-off-by: Devon Schwartz <devon.s.schwartz@utexas.edu>

Fixed reads so that they have proper parameters

Signed-off-by: Devon Schwartz <devon.s.schwartz@utexas.edu>

Fixed semicolon and variable instantiation

Signed-off-by: Devon Schwartz <devon.s.schwartz@utexas.edu>
```

using `git commit --amend`. You need to have only one `Signed-off-by` tag and one commit desciption. You can take some inspiration about how we format git commit messages from `git log`.

DevonSchwartz:
> Please, edit commit message:
> 
> ```
>  Combine last 3 commits into one commit that changes the macros
> 
> Changed the strcmp in lxcfs_read() to macros
> 
> Signed-off-by: Devon Schwartz <devon.s.schwartz@utexas.edu>
> 
> Fixed reads so that they have proper parameters
> 
> Signed-off-by: Devon Schwartz <devon.s.schwartz@utexas.edu>
> 
> Fixed semicolon and variable instantiation
> 
> Signed-off-by: Devon Schwartz <devon.s.schwartz@utexas.edu>
> ```
> 
> using `git commit --amend`. You need to have only one `Signed-off-by` tag and one commit desciption. You can take some inspiration about how we format git commit messages from `git log`.



> small nit

Thank you! Just changed the previous commit message to one sign-off and fixed the cgroup_is_enabled issue.

src/lxcfs.c
DevonSchwartz: Added change

DevonSchwartz:
I went ahead and added the checks to FUSE file system calls except open/opendir. Should we find a way to create a safe macro for the non-FUSE calls too, like rm()?

stgraber:
Yeah, a single clean commit would be good I think.

DevonSchwartz:
Should I resolve the change requested from above?

DevonSchwartz:
@mihalicyn Did I correctly address your desired changes? I think the merge is still blocked because of the review request on May 3.

mihalicyn:
Hey @DevonSchwartz 

sorry for a long delay with the reply.

Yeah, now it looks good.

For some reason GitHub runners get stuck to test this one. I guess you need to do something like `git commit --amend && git push -f` to retrigger tests.

DevonSchwartz:
Thanks @mihalicyn . Just made the push

PR #641, opened 2024-05-05
adding support for pthread_mutex_lock handling

sona78:
Updates:

- Standardized pthread_mutex_lock handling with replacement with interruptible locks for error handling
- Updated proc_loadavg.c with more robust handling to unlock mutexes that will cause errors

### DIFF
--- a/src/proc_loadavg.c
+++ b/src/proc_loadavg.c
@@ -138,7 +138,10 @@ static void insert_node(struct load_node **n, int locate)
 {
 	struct load_node *f;
 
-	pthread_mutex_lock(&load_hash[locate].lock);
+
+	if(mutex_lock_interruptible(&load_hash[locate].lock)){
+		return NULL;
+	}
 
 	/*
 	 * We have to recheck if the node we are looking for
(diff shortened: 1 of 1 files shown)


src/proc_loadavg.c
```diff
@@ -138,7 +138,10 @@ static void insert_node(struct load_node **n, int locate)
 {
 	struct load_node *f;
 
-	pthread_mutex_lock(&load_hash[locate].lock);
+
+	if(mutex_lock_interruptible(&load_hash[locate].lock)){
```
mihalicyn, line 142: what's the purpose replacing of a non-interruptible mutex with interruptible in this case?

Also, if you doing that, you need to handle "interrupt" case very-very carefully in all places from where the surrounding function (`insert_node`) is called.

src/proc_loadavg.c
```diff
@@ -138,7 +138,10 @@ static void insert_node(struct load_node **n, int locate)
...
+		return NULL;
```
mihalicyn, line 143: this function has a `void` return type. Have you tried to compile and test this code?

mihalicyn (changes requested):

mihalicyn:
a few things to consider:
- this code does not compile
- there is not Signed-off-by tag in the commit message
- please, explain a reason for this change

PR #661, opened 2024-10-02
lxcfs: fix readdir for procfs subtree

mihalicyn:
After #640 was merged we've got the entire
procfs subtree unavailable.

procfs can not be converted `LXCFS_TYPE_PROC` check in `readdir` because we don't have a designated `enum lxcfs_virt_t`-typed value for procfs directory. We can not introduce it too as it will break live upgrades of liblxcfs.

Fixes: #640


mihalicyn:
Added a test from https://github.com/lxc/lxcfs/pull/662

stgraber (approved):

PR #663, opened 2024-10-02
cgroups/cgfsng: improve swap accounting support detection

mihalicyn:
When LXCFS daemon runs in a root cgroup of cgroup2 tree, we need to go down the tree when checking for memory.swap.current.

We already have some logic to go up the tree from LXCFS daemon's cgroup, but it's useless when LXCFS daemon sits in the cgroup2 tree root.


mihalicyn:
Just for reference, these days it not possible to disable swap accounting on memcg:
https://github.com/torvalds/linux/commit/b25806dcd3d5248833f7d2544ee29a701735159f

mihalicyn:
cc @alexhudspith

brauner (approved):
This won't handle the case where no cgroup exists in `/sys/fs/cgroup/` at all but that's extremely unlikely so I'm not sure it's worth worrying about.

PR #665, opened 2024-11-13
proc_fuse: add psi(pressure stall information) procfs

loyou:
Kernel support psi(pressure stall information) since 4.20
with procfs /proc/pressure/{io,cpu,memory} and
cgroupv2 {io.pressure, cpu.pressure, memory.pressure}.

This patch add read-only psi procfs,
and people can get pressure information now.
Full functional feature for monitoring are still under investigation.


mihalicyn:
Hey @loyou,

thanks for working on this.

Please, notice that you have a failing test https://github.com/lxc/lxcfs/actions/runs/11809711932/job/32900481717?pr=665

src/bindings.h
```diff
@@ -57,6 +58,16 @@ enum lxcfs_virt_t {
 	LXC_TYPE_PROC_SLABINFO,
 #define LXC_TYPE_PROC_SLABINFO_PATH "/proc/slabinfo"
 
+	LXC_TYPE_PROC_PRESSURE,
```
mihalicyn, line 61: I don't think that you can add new items in the middle of `enum` as it can (and will) break liblxc live upgrades.

loyou: well, if add at the end, the LXCFS_TYPE_* macros will become really complicated,
i will think about a better way to deal with it.

loyou:
I'm a little confused with the test result,
```
==> Testing /sys/devices/system/cpu/online
../src/lxcfs.c: 910: lxcfs_read: unknown file type: path=/sys/devices/system/cpu/online, type=-1, fi->fh=105965433156896
cat: /tmp/tmp.tDXXvo8T0e/sys/devices/system/cpu/online: Invalid argument
../src/lxcfs.c: 988: lxcfs_release: unknown file type: path=/sys/devices/system/cpu/online, type=-1, fi->fh=105965433156896
```
the failed line number does not match the patched file lines,
and `cat /sys/devices/system/cpu/online` in lxc container with patched lxcfs, it works fine.
I will resolve the live upgrade issue first and then check it again.

src/bindings.h
loyou: It become complicated after commit cf02ccbe, when new proc type is added, we have to change the macro.
Maybe revert to original way is better. I rebased the code and force pushed, please kindly review it again.
Thanks.

src/proc_fuse.c
```diff
@@ -164,17 +184,30 @@ __lxcfs_fuse_ops int proc_readdir(const char *path, void *buf,
...
+		    dir_filler(filler, buf, "..",		0) != 0 ||
+		    dir_filler(filler, buf, "cpuinfo",		0) != 0 ||
+		    dir_filler(filler, buf, "meminfo",		0) != 0 ||
+		    dir_filler(filler, buf, "stat",		0) != 0 ||
+		    dir_filler(filler, buf, "uptime",		0) != 0 ||
+		    dir_filler(filler, buf, "diskstats",	0) != 0 ||
+		    dir_filler(filler, buf, "swaps",		0) != 0 ||
+		    dir_filler(filler, buf, "loadavg",		0) != 0 ||
+		    dir_filler(filler, buf, "slabinfo",		0) != 0 ||
+		    dirent_filler(filler, path, "pressure", buf, 0) != 0)
```
hallyn, line 198: I haven't tested this, but it looks like if the host does not have /proc/pressure, lxcfs will have a /proc/pressure though it will be empty.  I don't know whether that matters, especially as noone directly mounts /var/lib/lxcfs/proc into the container.

hallyn (approved):
Looks good to me.

hallyn:
Will wait briefly to see whether @mihalicyn had any more comments before merging.

stgraber:
Going to do a quick rebase on this one so it triggers the new tests we have.

mihalicyn (approved):
LGTM!

Issue #670, opened 2025-01-05
slab_reclaimable from memory.stat should be added to meminfo's MemAvailable


asainkujovic:
The template below is mostly useful for bug reports and support questions.
Feel free to remove anything which doesn't apply to you and add more information where it makes sense.

# Required information

 * Distribution and lxcfs version: Any, with active lxcfs mounts inside container (focus on /proc/meminfo)

# Issue description

Inside container, free memory is bigger than it would be in classic hosts and bare-metal systems in same situations.
Slab memory grows and grows after using the system, which is normal, that memory can be reclaimed and will be cleared in favor of new requests. But LXC container in the end shows that as USED memory which is different of regular hosts behaviour. Regular behaviour is that Slab-reclaimable memory is a part of Available (MemAvail) value. It is not totally free (as not used at all) but it is available for user and system when it needs it.

# Steps to reproduce

 1. Prepare some directory where there is significatnt number of files (10GB or more for example)
 2. run `watch 'free -m ; head /proc/meminfo` # free is based on /proc/meminfo which is created by lxcfs based on memory.stat
 3. clear all buffer/slab memory with: echo 3 | sudo tee /proc/sys/vm/drop_caches
 4. run du -hd1 /directory/path # in other terminal, this will cause growth of slab memory
 5. inspect growth of values, USED memory rise up for significat value (available memory drops for significant value that is equal to sreclaimable plus more), and it should not.

# Information to attach

I am not an expert about all details about memory, meminfo, memstat, memory.current, buffers, caches, inodes, active_file, nonactive_file ... all of this is very complex behind, even read somewhere that free -m calculates things diferent in different environments/linux flavors. But:
1) I had a concrete problem:
- LXC in container shows more and more usage and send me ALARMS about it
- and in reality no real memory is taken, but memory that will be cleared on request (tested well, slab-reclaimable clears itself in favor of for example test memory request: `dd if=/dev/zero bs=128M count=10 status=progress | tail | sleep 25`
2) I tested this problem on various systems well in least 15 days with helper scripts that shows growth/change of values
3) After all struggle and testing and complexity, everything lead to simple conclusion:
**slab_reclaimable should be added to MemAvailable** for sure

[...]

PR #671, opened 2025-01-05
meminfo: Add slab_reclaimable to MemAvailable

asainkujovic:
This PR proposal is created for the problem explained at [issue 670](https://github.com/lxc/lxcfs/issues/670)

Slab-Reclaimable memory in any system should not be part of "used" but oposite, should be part of "available" memory.

### DIFF
--- a/src/proc_fuse.c
+++ b/src/proc_fuse.c
@@ -63,6 +63,9 @@ struct memory_stat {
 	uint64_t total_inactive_file;
 	uint64_t total_active_file;
 	uint64_t total_unevictable;
+	uint64_t slab;
+	uint64_t slab_reclaimable;
+	uint64_t slab_unreclaimable;
 };
 
 static off_t get_procfile_size(const char *path)
@@ -1293,6 +1296,12 @@ static bool cgroup_parse_memory_stat(const char *cgroup, struct memory_stat *mst
 			sscanf(line, unified ? "active_file %" PRIu64 : "total_active_file %" PRIu64, &(mstat->total_active_file));
 		} else if (startswith(line, unified ? "unevictable" : "total_unevictable")) {
 			sscanf(line, unified ? "unevictable %" PRIu64 : "total_unevictable %" PRIu64, &(mstat->total_unevictable));
+		} else if (unified && startswith(line, "slab ")) {
+			sscanf(line, "slab %" PRIu64, &(mstat->slab));
+		} else if (unified && startswith(line, "slab_reclaimable")) {
+			sscanf(line, "slab_reclaimable %" PRIu64, &(mstat->slab_reclaimable));
+		} else if (unified && startswith(line, "slab_unreclaimable")) {
+			sscanf(line, "slab_unreclaimable %" PRIu64, &(mstat->slab_unreclaimable));
 		}
 	}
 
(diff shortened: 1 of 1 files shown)


src/proc_fuse.c
```diff
@@ -1389,7 +1398,8 @@ static int proc_meminfo_read(char *buf, size_t size, off_t offset,
 			snprintf(lbuf, 100, "MemFree:        %8" PRIu64 " kB\n", memlimit - memusage);
 			printme = lbuf;
 		} else if (startswith(line, "MemAvailable:")) {
-			snprintf(lbuf, 100, "MemAvailable:   %8" PRIu64 " kB\n", memlimit - memusage + (mstat.total_active_file + mstat.total_inactive_file) / 1024);
+			snprintf(lbuf, 100, "MemAvailable:   %8" PRIu64 " kB\n", memlimit - memusage +
+				(mstat.total_active_file + mstat.total_inactive_file + mstat.slab_reclaimable) / 1024);
```
mihalicyn, line 1402: I believe this approach is good enough. For reference: https://github.com/torvalds/linux/blob/13563da6ffcf49b8b45772e40b35f96926a7ee1e/mm/show_mem.c#L64

src/proc_fuse.c
```diff
@@ -1293,6 +1296,12 @@ static bool cgroup_parse_memory_stat(const char *cgroup, struct memory_stat *mst
 			sscanf(line, unified ? "active_file %" PRIu64 : "total_active_file %" PRIu64, &(mstat->total_active_file));
 		} else if (startswith(line, unified ? "unevictable" : "total_unevictable")) {
 			sscanf(line, unified ? "unevictable %" PRIu64 : "total_unevictable %" PRIu64, &(mstat->total_unevictable));
+		} else if (unified && startswith(line, "slab ")) {
```
mihalicyn, line 1299: nit: unnecessary space after `slab`.

mihalicyn (approved):
LGTM.

Hi @asainkujovic,

thanks for your work on this one!

Kind regards,
Alex

src/proc_fuse.c
asainkujovic: sorry, i added that space, so that bellow 'slab_reclaimable' and 'slab_unreclaimable' do not match with 'starts_with'. So that space is needed or can be removed but then line "slab" must be below other two lines (if/else) ... which could be more confusing, someone could bring it up again...

asainkujovic:
This PR as it is should be better option than current situation. And I still proposing it as it will solve mentioned issue.
However, full story (as you probably already know) is more complicated, Kernel changed the way of calculations, and newest kernel code use zones, and watermark and half estimation where half of reclaimable values stays as reserved. So for completeness i would sum up my view of situation. If we imagine some init pseudocode:
```
    # variables for MemAvail calculation:
    memlimit        = file_contents(/sys/fs/cgroup/memory.max)
    memusage        = file_contents(/sys/fs/cgroup/memory.current)
    // watermark inside lxc container could be read from memory.peak file, and kernel calculates it by looping through ZONES
    watermark       = file_contents(/sys/fs/cgrop/memory.peak IF that file exists ELSE 0)
    slab_recl       = file_line_val(/sys/fs/cgroup/memory.stat, line:slab_reclaimable)
    total_file      = file_line_val(/sys/fs/cgroup/memory.stat, line:active_file)
                     +file_line_val(/sys/fs/cgroup/memory.stat, line:inactive_file)

    estimated_file  = total_file - min(total_file/2, watermark)
    estimated_srecl = slab_recl  - min(slab_recl /2, watermark)
    
    // (1) current kernel code calculates someting like this:
    MemAvail = memlimit - memusage + estimated_file + estimated_srecl

    // (2) current LXCFS version calculates MemAvail like this:
    MemAvail = memlimit - memusage + total_file

    // (3) this PR of lxcfs calculates it like this:
    MemAvail = memlimit - memusage + total_file + slab_recl
```

So solution (3) is still better than (2) and I am still proposing this PR as solution (2), but more complex way (1) could also be written in some next PR, just for now I am not so competent to predict possible problems and to be 100% sure (i could be just 75% sure for example). Watermark for cgroup should be exposed from 2022 as memory.peak file, but not sure if that is the same as 'low_watermark' that is used in Kernel calculations. But all of this is close related to memory problems that I have to inspect and work on anyway, and I can inform if some better improvement / proposal comes up from me or my colleagues.

Issue #678, opened 2025-03-17
Loadavg calculation inside containers stops for about 70 min after live migration when running on a GCP VM


DeyanSG:
When running lxcfs inside a GCP VM, we encountered a strange situation after the VM was live-migrated to a different host.

It seems that the kernel missed some interrupts due to the live migration, which caused the thread responsible for calculating the loadavg values inside containers to fail to wake up. This resulted in a freeze in the loadavg values for about 70 minutes.

Initially, when we observed the issue, we decided to try reloading lxcfs, hoping this would restart the loadavg thread. However, instead of resolving the issue, all attempts to read the files managed by lxcfs ended up hanging in the read call, and the reload did not occur. We had to restart the service and all containers running on the VM to resume normal operations.

A kernel backtrace of the thread before the restart showed that it was sleeping:

```
crash> bt 4467
PID: 4467     TASK: ffff88a052bf8000  CPU: 43   COMMAND: "lxcfs"
 #0 [ffffc9000eea7d90] __schedule at ffffffff81bd7161
 #1 [ffffc9000eea7e38] schedule at ffffffff81bd785f
 #2 [ffffc9000eea7e58] do_nanosleep at ffffffff81bddcae
 #3 [ffffc9000eea7ea0] hrtimer_nanosleep_restart at ffffffff81bdde26
 #4 [ffffc9000eea7f18] __x64_sys_restart_syscall at ffffffff810c05df
 #5 [ffffc9000eea7f28] x64_sys_call at ffffffff81003316
 #6 [ffffc9000eea7f38] do_syscall_64 at ffffffff81bc2a35
 #7 [ffffc9000eea7f50] entry_SYSCALL_64_after_hwframe at ffffffff81c000af
    RIP: 00007fa4742d4355  RSP: 00007fa473fffd80  RFLAGS: 00000293
    RAX: ffffffffffffffda  RBX: 00007fa474000640  RCX: 00007fa4742d4355
    RDX: 00007fa473fffdc0  RSI: 0000000000000000  RDI: 0000000000000000
    RBP: 00007fa473fffe50   R8: 0000000000000000   R9: 0000000000000000
    R10: 0000000000000000  R11: 0000000000000293  R12: 0000000000000001
    R13: 0000000000000000  R14: 00007fa474289bc0  R15: 0000000000000000
    ORIG_RAX: 00000000000000db  CS: 0033  SS: 002b
crash>
```

[...]

PR #679, opened 2025-03-17
proc_loadavg: Prevent integer overflow calculating the sleep interval

DeyanSG:
This resolves the issue with reload described in https://github.com/lxc/lxcfs/issues/678

### DIFF
--- a/src/proc_loadavg.c
+++ b/src/proc_loadavg.c
@@ -504,6 +504,7 @@ static void *load_begin(void *arg)
 	int first_node, sum;
 	struct load_node *f;
 	clock_t time1, time2;
+	int sleep_time;
 
 	for (;;) {
 		if (loadavg_stop == 1)
@@ -542,8 +543,9 @@ static void *load_begin(void *arg)
 			return NULL;
 
 		time2 = clock();
-		usleep(FLUSH_TIME * 1000000 -
-		       (int)((time2 - time1) * 1000000 / CLOCKS_PER_SEC));
+		sleep_time = FLUSH_TIME - (int)((time2 - time1) / CLOCKS_PER_SEC);
+		if ((sleep_time > 0) && (sleep_time <= FLUSH_TIME))
+			usleep(sleep_time * 1000000);
 	}
 }
 


Blub:
The signal handler will set `need_reload=1`, which is a bit of a weird side effect.
It *should* be okay, since when we reach this code, we're either in `do_reload()` where that's already the case, or in `main()` during shutdown, but personally I don't like depending on this kind of knowledge, as it is a maintenance burden.
(I wonder if we should introduce a more specific signal there (`SIGALRM` is common to wake up sleeps I think, or some RT signal)

mihalicyn:
Hello @DeyanSG,

first of all, thanks for your report and proposed fix!

From #678:
>This led us to conclude that we missed the wake-up time for the timer started by the usleep in the loadavg thread. We calculated that it would take about 70 minutes for the timer to overflow and reach the same value again, allowing normal operations to continue.

you are referring to this part of code:
```
	clock_t time1, time2;

	for (;;) {
		if (loadavg_stop == 1)
			return NULL;

		time1 = clock();
<...>
		if (loadavg_stop == 1)
			return NULL;

		time2 = clock();
		usleep(FLUSH_TIME * 1000000 -
		       (int)((time2 - time1) * 1000000 / CLOCKS_PER_SEC));
```
right?

Please, can you test the following simple fix:
```
--- a/src/proc_loadavg.c
+++ b/src/proc_loadavg.c
@@ -504,6 +504,7 @@ static void *load_begin(void *arg)
        int first_node, sum;
        struct load_node *f;
        clock_t time1, time2;
+       int sleep_time;
 
        for (;;) {
                if (loadavg_stop == 1)
@@ -542,8 +543,9 @@ static void *load_begin(void *arg)
                        return NULL;
 
                time2 = clock();
-               usleep(FLUSH_TIME * 1000000 -
-                      (int)((time2 - time1) * 1000000 / CLOCKS_PER_SEC));
+               sleep_time = FLUSH_TIME - (int)((time2 - time1) / CLOCKS_PER_SEC);
+               if ((sleep_time > 0) && (sleep_time <= FLUSH_TIME))
+                       usleep(sleep_time * 1000000);
        }
 }
``` 

Kind regards,
Alex

Blub:
Ah yes, catching the overflow definitely makes sense. We'd still be a bit delayed, but that's much less problematic than a near-endless sleep call and can still be addressed separately.

DeyanSG:
Hi @mihalicyn and @Blub,

Thanks for looking into this.

It seems I somehow missed the overflow in the code and went on to look into the kernel code. Thanks for spotting this, @mihalicyn. We will deploy a patched version today that includes the suggested fix and some additional logging in case we skip the sleep, so we can see all the values.

The issue is tricky to reproduce, as we do not trigger these live migrations. They are part of some GCE maintenance, and so far, the issue during a migration seems to occur rarely. Nevertheless, we will do our best to test and see if we can determine whether this patch will solve the issue.

@Blub, I tested with SIGALRM and SIGVTALRM. However, sending these to the thread seems to kill the whole process. I am also not sure how good of an idea it is to set a custom handler for these signals. I also tested with SIGUSR2, and even without installing a handler for this signal, it seems to work correctly. Let me know if you are keen on merging this with SIGUSR2 (or another signal), or if we will stick to just the solution provided by @mihalicyn if it works.

Regards,
Deyan

Blub:
`SIGUSR2` is already used up by existing lxcfs code.

I think installing a handler for `SIGALRM` should be fine, we can choose ourselves how such signals should be dealt with, and `SIGALRM` is commonly for timeouts (eg. using the `alarm(2)` call which is meant to interrupt (without killing the process). (Eg. there are no file-locking methods with a timeout, so the only way to try such locks *with* a time out is to set a timer or alarm or emulate such a thing via threads+sleep)).
If we ever need *that* behavior we'd want to use timers (`timer_create(2)`) instead of `alarm(2)` anyway, as with timers you can target specific threads.

DeyanSG:
Hi @Blub ,

I've modified it to use `SIGALRM` as suggested and did a quick test. It seems to work fine this way.

Feel free to suggest more improvements.

Regards,
Deyan

DeyanSG:
Hi, @mihalicyn!

I can confirm that the changes proposed by you seem to fix the issue. Here are some logs that I have gathered. All cases can be mapped to a GCE VM live migration happening at the same time. There are even cases in which we've skipped 2-3 consecutive sleeps that could have resulted in this behavior:

```
Mar 23 11:55:02 HostA lxcfs[4504]: ../src/proc_loadavg.c: 550: load_begin: Skipped sleep -1s (4293967296)
Mar 23 12:00:30 HostB lxcfs[4487]: ../src/proc_loadavg.c: 550: load_begin: Skipped sleep -16s (4278967296)
Mar 23 12:10:03 HostC lxcfs[4509]: ../src/proc_loadavg.c: 550: load_begin: Skipped sleep -57s (4237967296)
Mar 23 12:25:35 HostD lxcfs[4429]: ../src/proc_loadavg.c: 550: load_begin: Skipped sleep -14s (4280967296)
Mar 23 12:35:11 HostE lxcfs[4492]: ../src/proc_loadavg.c: 550: load_begin: Skipped sleep -12s (4282967296)
Mar 23 12:40:12 HostF lxcfs[4499]: ../src/proc_loadavg.c: 550: load_begin: Skipped sleep -2s (4292967296)
Mar 23 13:14:47 HostG lxcfs[4502]: ../src/proc_loadavg.c: 550: load_begin: Skipped sleep -1s (4293967296)
Mar 23 13:14:52 HostG lxcfs[4502]: ../src/proc_loadavg.c: 550: load_begin: Skipped sleep -2s (4292967296)
Mar 23 14:22:59 HostH lxcfs[4495]: ../src/proc_loadavg.c: 550: load_begin: Skipped sleep -6s (4288967296)
Mar 23 14:23:03 HostH lxcfs[4495]: ../src/proc_loadavg.c: 550: load_begin: Skipped sleep 0s (0)
Mar 23 14:27:58 HostI lxcfs[4452]: ../src/proc_loadavg.c: 550: load_begin: Skipped sleep -6s (4288967296)
Mar 23 14:28:18 HostI lxcfs[4452]: ../src/proc_loadavg.c: 550: load_begin: Skipped sleep -8s (4286967296)
Mar 23 14:28:23 HostI lxcfs[4452]: ../src/proc_loadavg.c: 550: load_begin: Skipped sleep -3s (4291967296)
Mar 23 14:32:41 HostJ lxcfs[4446]: ../src/proc_loadavg.c: 550: load_begin: Skipped sleep -2s (4292967296)                                                                                                                                     
```

After reviewing our monitoring data, I can confirm that we did not experience any significant freezes in loadavg values within the containers during these events.

I've committed your code changes as suggested and updated the MR accordingly.

Thank you for your efforts in addressing this issue!

Regards,
Deyan

mihalicyn (approved):
LGTM.

Thanks!

PR #686, opened 2025-08-22
Add zswap accounting

kadinsayani:
Adds support for reporting zswap memory statistics in /proc/meminfo.


kadinsayani:
@mihalicyn I've kept this as draft for now since it doesn't include any checks for kernel zswap support.

mihalicyn:
We've discussed this PR privately with Kadin and he will post a new version soon.

kadinsayani:
Ready for review :white_check_mark:

src/bindings.h
```diff
@@ -120,6 +120,7 @@ struct lxcfs_opts {
 	bool swap_off;
 	bool use_pidfd;
 	bool use_cfs;
+	bool zswap_off;
```
mihalicyn, line 123: please, move this new field at the end of the structure. This is important, because
this structure is a part of API convention between libLXCFS and LXCFS main binary.
LXCFS allows for live libLXCFS upgrades, and `struct lxcfs_opts` allocated from the main binary (see https://github.com/lxc/lxcfs/blob/e2973d1e38c89aa9e6b4fad1a97ed524076c73fa/src/lxcfs.c#L1361)
which means that you can go to a newer version of libLXCFS that will assume a new `struct lxcfs_opts` layout, while the memory for it was allocated with the old structure in mind. It means that on libLXCFS side, before accessing any newly added fields, we need to consult with `->version` field of this struct.

So the right thing to do is:
- only add new members at the end of this structure
- bump structure version https://github.com/lxc/lxcfs/blob/e2973d1e38c89aa9e6b4fad1a97ed524076c73fa/src/lxcfs.c#L1370
- before accessing `zswap_off` field, ensure that `->version >= 3`.

mihalicyn: And yes, this is bad that our tests can't catch it. I need to think about how we can improve our tests in there. But this can be quite challenging.

src/cgroups/cgfsng.c
```diff
@@ -633,75 +633,18 @@ static int cgfsng_get_memory_slabinfo_fd(struct cgroup_ops *ops, const char *cgr
...
-			}
-		}
-	} else {
-		/*
-		 * We can check a (file) presence on the current
-		 * level and go up in the cgroup tree if needed.
-		 */
-		ret = cgroup_walkup_to_root(ops->cgroup2_root_fd, h->fd, cgroup_rel, file, &junk_value) == 0;
-	}
+	fd = openat(h->fd, ".", O_DIRECTORY | O_RDONLY | O_CLOEXEC);
```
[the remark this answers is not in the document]
kadinsayani: > Aren't we leaking this fd? I can't see where we close it.

Nice catch, that `fd` was getting leaked. 

> IMHO, this helper function signature should looks like

Sounds good, thanks for the suggestion :)

src/bindings.h
kadinsayani: Thanks, I've made those changes.

Issue #654, opened 2024-07-12
/proc/stat processes, procs_running,procs_blocked same for host and container


kartoffelheinz:
# Required information

 * Distribution:
   * PRETTY_NAME="Debian GNU/Linux 11 (bullseye)"
NAME="Debian GNU/Linux"
VERSION_ID="11"
VERSION="11 (bullseye)"
VERSION_CODENAME=bullseye
ID=debian
HOME_URL="https://www.debian.org/"
SUPPORT_URL="https://www.debian.org/support"
BUG_REPORT_URL="https://bugs.debian.org/"

 * LXCFS version: 6.0.1
 * The output of
   * `uname -a: Linux rmltest 5.10.0-30-amd64 #1 SMP Debian 5.10.218-1 (2024-06-01) x86_64 GNU/Linux`
   * `
sysfs /sys sysfs rw,nosuid,nodev,noexec,relatime 0 0
proc /proc proc rw,nosuid,nodev,noexec,relatime,gid=4,hidepid=invisible 0 0
udev /dev devtmpfs rw,nosuid,relatime,size=4043224k,nr_inodes=1010806,mode=755 0 0
devpts /dev/pts devpts rw,nosuid,noexec,relatime,gid=5,mode=620,ptmxmode=000 0 0
tmpfs /run tmpfs rw,nosuid,nodev,noexec,relatime,size=813616k,mode=755 0 0
/dev/mapper/system-system_root / xfs rw,relatime,attr2,inode64,logbufs=8,logbsize=32k,noquota 0 0
securityfs /sys/kernel/security securityfs rw,nosuid,nodev,noexec,relatime 0 0
tmpfs /dev/shm tmpfs rw,nosuid,nodev 0 0
tmpfs /run/lock tmpfs rw,nosuid,nodev,noexec,relatime,size=5120k 0 0
tmpfs /sys/fs/cgroup tmpfs ro,nosuid,nodev,noexec,size=4096k,nr_inodes=1024,mode=755 0 0
cgroup2 /sys/fs/cgroup/unified cgroup2 rw,nosuid,nodev,noexec,relatime 0 0
cgroup /sys/fs/cgroup/systemd cgroup rw,nosuid,nodev,noexec,relatime,xattr,name=systemd 0 0
pstore /sys/fs/pstore pstore rw,nosuid,nodev,noexec,relatime 0 0
none /sys/fs/bpf bpf rw,nosuid,nodev,noexec,relatime,mode=700 0 0
[...]

Issue #593, opened 2023-04-04
/proc/stat is not work in cgroup v2


hdp7891000:
Including the latest 5.0.3 version……

hdp7891000:
> 
Thanks.
I use **cgroup v2** on the host, and use docker's **-v /var/lib/lxcfs/proc/stat:/proc/stat** option to bind /proc/stat to the container.
In the container, **cat /proc/stat** gets the contents of the host.

mjrodan1:
I am also using lxcfs with admission webhook. I have same problem. 
[https://github.com/ymping/lxcfs-admission-webhook](url)

in this project, this code normally mounts
`
  - /var/lib/lxc/lxcfs/proc/cpuinfo:/proc/cpuinfo
  - /var/lib/lxc/lxcfs/proc/diskstats:/proc/diskstats
  - /var/lib/lxc/lxcfs/proc/meminfo:/proc/meminfo
  - /var/lib/lxc/lxcfs/proc/stat:/proc/stat
  - /var/lib/lxc/lxcfs/proc/swaps:/proc/swaps
  - /var/lib/lxc/lxcfs/proc/uptime:/proc/uptime
  - /var/lib/lxc/lxcfs/proc/loadavg:/proc/loadavg
  - /var/lib/lxc/lxcfs/sys/devices/system/cpu/online:/sys/devices/system/cpu/online
  - `

so I added code to mount 
`
- /var/lib/lxcfs/proc/slabinfo:/proc/slabinfo:rw \
- /var/lib/lxcfs/sys/devices/system/cpu:/sys/devices/system/cpu:rw 
`
(changed cpu mount point /cpu/online to /cpu)

I could see all the cpu informations when I see /proc/stat.
But I could check /proc/cpuinfo works normally working

mihalicyn:
@hdp7891000

Couldn't you show the lxcfs logs?

@mjrodan1 

> I could see all the cpu informations when I see /proc/stat.
> But I could check /proc/cpuinfo works normally working

Couldn't you show lxcfs logs and contents of both files from inside the container?

hdp7891000:
lxcfs doesn't print anything special.

I think read_cpuacct_usage_all does not support cgroup v2, because there is no cpuacct.usage_all or cpuacct.usage_percpu.
```
        /*
	 * Read cpuacct.usage_all for all CPUs.
	 * If the cpuacct cgroup is present, it is used to calculate the container's
[...]

PR #687, opened 2025-09-01
proc_fuse: fix proc_stat_read reporting host cpu count under cgroup v2

last-las:
cgroup v2 doesn't have a file similar to `cpuacct.usage_all`. Therefore, the `proc_stat_read` in cgroup v2 always falls back to the host's /proc/stat. As a result, the CPU count reported by this file is incorrect, e.g. issues https://github.com/lxc/lxcfs/issues/593 and https://github.com/lxc/lxcfs/issues/654.

However, the similar `proc_cpuinfo_read` function does report the right CPU counts. It works by filtering the host's file using both `cpuset.cpus` and `cpu.max`. This patch fixes `proc_stat_read` by using the same method.


stgraber:
@last-las can you edit your commit message to include the required Signed-off-by line? (see CONTRIBUTING.md)

last-las:
> @last-las can you edit your commit message to include the required Signed-off-by line? (see CONTRIBUTING.md)

done!

stgraber:

mihalicyn:
Hi @last-las,

thanks for working on this and preparing a PR.

I was thinking about this way too, before. But my concern was and is that if we handle this like that, what we end up with is incorrect `/proc/stat` statistics anyways. Yes, CPUs number will be correct with this applied, but all the parameters:
```
			   "%*s"        /* <skip> */
			   " %" PRIu64  /* user */
			   " %" PRIu64  /* nice */
			   " %" PRIu64  /* system */
			   " %" PRIu64  /* idle */
			   " %" PRIu64  /* iowait */
			   " %" PRIu64  /* irq */
			   " %" PRIu64  /* softirq */
			   " %" PRIu64  /* steal */
			   " %" PRIu64  /* guest */
			   " %" PRIu64, /* guest_nice */
```
will still be taken from a host, isn't it?

So the question is then should we instead go and make a proper kernel support for that CPU time accounting (as it was in cgroupv1) or just apply this half-measure?

Could you elaborate why this cpu-filtering-only solution even works for you? What kind of use-case do you have for this `/proc/stat` thing inside the container? For me it looks like we instead of giving a user a wrong information now will be giving the user another wrong information. :)

mihalicyn (approved):
In general, if it helps users I see no problem merging this.

PR #690, opened 2025-10-15
lxcfs: fix wrong cpu count when setting cfs in hierarchy

### LINKED ISSUE #688: Wrong cpu number reported, when only CFS (cpu.max) cgroup2 limit is set
cpu count was read in current cgroup only.
if parent cgroup limit cpu to 2, but current cgroup limit is unlimit, 
lxcfs will show all cpu for container.

okhowang:
Resolve #688

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -167,6 +167,7 @@ if not libfuse.found()
 endif
 
 libdl = cc.find_library('dl')
+libm = cc.find_library('m')
 threads = dependency('threads')
 
 config_h = configure_file(
@@ -221,6 +222,7 @@ liblxcfs_common_dependencies = declare_dependency(
     sources: liblxcfs_sources,
     dependencies: [
         threads,
+        libm,
         libfuse,
     ])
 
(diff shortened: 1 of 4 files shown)


stgraber:

mihalicyn (changes requested):
Please, reply on my comment in https://github.com/lxc/lxcfs/issues/688#issuecomment-3840029800

mihalicyn (changes requested):
Thanks for preparing and submitting this.

I believe that this is on the right direction but we need to slightly rework the code.

Also, please look https://github.com/lxc/lxcfs/issues/688#issuecomment-4090064168

Kind regards,
Alex

mihalicyn (changes requested):

mihalicyn (changes requested):

mihalicyn (changes requested):
rebase is needed

okhowang:
> Hi @okhowang,
> 
> yes, please remove the test you've added, because the version we have in `main` is more advanced and covers this stuff better.

test commit removed

mihalicyn (approved):
LGTM

Thanks, @okhowang !

PR #692, opened 2025-11-18
fix: reuse cgroup2 superblock options when mount cgroupfs

xujihui1985:
the cgroup2 mount in __cg_mount_direct() previously ignored the superblock options applied on the existing cgroup2 mountpoint. This caused the newly created cgroup2 mount to miss important options such as nsdelegate, memory_localevents, and memory_recursiveprot, which may affect controller behavior and result in inconsistent semantics across mount points.

This patch introduces extract_cgroup2_super_opts() to locate the first relevant cgroup2 superblock option in the existing mount options string, and get_mount_opts() to retrieve the mount options for a given target and filesystem type from /proc/self/mounts. These helpers allow the new mount to inherit the same superblock configuration as the system's primary cgroup2 mount.

### DIFF
--- a/src/cgroups/cgfsng.c
+++ b/src/cgroups/cgfsng.c
@@ -399,30 +399,30 @@ static void trim(char *s)
  */
 static int __cg_mount_direct(struct hierarchy *h, const char *controllerpath)
 {
-	__do_free char *controllers = NULL;
-	__do_free const char *opts = NULL;
-	char *fstype = "cgroup2";
-	unsigned long flags = 0;
-	int ret;
+	 __do_free char *controllers = NULL;
+	 __do_free const char *opts = NULL;
+	 char *fstype = "cgroup2";
+	 unsigned long flags = 0;
+	 int ret;
 
-	flags |= MS_NOSUID;
-	flags |= MS_NOEXEC;
-	flags |= MS_NODEV;
-	flags |= MS_RELATIME;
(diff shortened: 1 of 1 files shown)


xujihui1985:
Hi @stgraber and @mihalicyn,
I was wondering if we could simply bind-mount the cgroupfs into the cgroup controller path. instread of mount again and pick the mount options from cgroupfs, since mount options like nsdelegate and memory_recursiveprot are global under the initial cgroup namespace, even in the separate mountns, so if we mount with a different options, that actually change the options on the root cgroupfs

src/cgroups/cgfsng.c
```diff
@@ -399,27 +399,35 @@ static void trim(char *s)
...
+		}
+		const char *super_opts = extract_cgroup2_super_opts(opts);
+		ret = mount(fstype, controllerpath, fstype, flags, super_opts);
 	}
 
-	ret = mount("cgroup", controllerpath, fstype, flags, controllers);
 	if (ret < 0)
 		return -1;
-
+
```
mihalicyn: excess whitespace

xujihui1985:
> I like the approach. Please can you fix whitespaces to match our codestyle?

thanks so much for the reply, I will fix the review suggestion ASAP

src/cgroups/cgfsng.c
xujihui1985: fixed

PR #696, opened 2025-12-25
feat: add --force-render-cgroup args to make lxcfs render in view of specific cgroup

okhowang:
it's useful in some complex container hierarchy,
many container work in same parent cgroup for sharing resource (cpu and other)
but need inspect total usage of parent cgroup in single container


mihalicyn (changes requested):
I've left a comment before.

PR #726, opened 2026-05-11
lxcfs: : Don't register .poll at all unless --enable-psi-poll is passed

tomponline:
Without .poll in lxcfs_ops, the kernel handles poll entirely in-kernel (returning DEFAULT_POLLMASK without ever contacting userspace), so there is no FUSE round-trip and no D state when LXCFS is suspended.

However this only fixes the problem as long as `--enable-psi-poll` is not passed.

Partial fix for https://github.com/canonical/lxd/issues/17983

### DIFF
--- a/src/lxcfs.c
+++ b/src/lxcfs.c
@@ -700,7 +700,7 @@ static void *lxcfs_init(struct fuse_conn_info *conn, struct fuse_config *cfg)
 	return fuse_get_context()->private_data;
 }
 
-const struct fuse_operations lxcfs_ops = {
+struct fuse_operations lxcfs_ops = {
 	.access		= lxcfs_access,
 	.chmod		= lxcfs_chmod,
 	.chown		= lxcfs_chown,
@@ -719,7 +719,6 @@ const struct fuse_operations lxcfs_ops = {
 	.truncate	= lxcfs_truncate,
 	.write		= lxcfs_write,
 	.readlink	= lxcfs_readlink,
-	.poll		= lxcfs_poll,
 
 	.create		= NULL,
 	.destroy	= NULL,
@@ -1051,6 +1050,9 @@ int main(int argc, char *argv[])
 
 	if (load_use && start_loadavg() != 0)
 		goto out;
+
+	if (opts->psi_poll_on)
+		lxcfs_ops.poll = lxcfs_poll;
 
 	if (!fuse_main(fuse_argc, fuse_argv, &lxcfs_ops, opts))
 		ret = EXIT_SUCCESS;


tomponline:
Hi @mihalicyn what are your thoughts on this one?

I realise this is only really a band-aid solution, so I wondered if you had a more full-proof solution?

Thanks!

mihalicyn:
Hi Tom,

thank you for the patch. WDYT about the following:
```
diff --git a/src/lxcfs.c b/src/lxcfs.c
index 9ecdffd..5fb3360 100644
--- a/src/lxcfs.c
+++ b/src/lxcfs.c
@@ -540,9 +540,24 @@ int lxcfs_write(const char *path, const char *buf, size_t size, off_t offset,
 int lxcfs_poll(const char *path, struct fuse_file_info *fi,
               struct fuse_pollhandle *ph, unsigned *reventsp)
 {
+       struct fuse_context *fc = fuse_get_context();
+       struct lxcfs_opts *opts = fc->private_data;
        int ret;
        enum lxcfs_virt_t type;
 
+       /*
+        * Note, that if fuse daemon returns ENOSYS once, then
+        * kernel caches this result in fm->fc->no_poll and never
+        * sends FUSE_POLL again. It is not a problem for us now,
+        * because we only need poll() for PSI.
+        *
+        * See #726 on GitHub.
+        */
+       if (!opts->psi_poll_on) {
+               fuse_pollhandle_destroy(ph);
+               return -ENOSYS;
+       }
+
        type = file_info_type(fi);
 
        if (LXCFS_TYPE_PROC(type)) {
```

mihalicyn:
> I realise this is only really a band-aid solution, so I wondered if you had a more full-proof solution?

At this point no. I need to dive into suspend code to figure out why it can't just cancel request. A long ago I've added this https://github.com/lxc/lxcfs/pull/594 to deal with cases like this one and avoid D-states when possible.

tomponline:
> Hi Tom,
> 
> thank you for the patch. WDYT about the following:
> 
> ```
> diff --git a/src/lxcfs.c b/src/lxcfs.c
> index 9ecdffd..5fb3360 100644
> --- a/src/lxcfs.c
> +++ b/src/lxcfs.c
> @@ -540,9 +540,24 @@ int lxcfs_write(const char *path, const char *buf, size_t size, off_t offset,
>  int lxcfs_poll(const char *path, struct fuse_file_info *fi,
>                struct fuse_pollhandle *ph, unsigned *reventsp)
>  {
> +       struct fuse_context *fc = fuse_get_context();
> +       struct lxcfs_opts *opts = fc->private_data;
>         int ret;
>         enum lxcfs_virt_t type;
>  
> +       /*
> +        * Note, that if fuse daemon returns ENOSYS once, then
> +        * kernel caches this result in fm->fc->no_poll and never
> +        * sends FUSE_POLL again. It is not a problem for us now,
> +        * because we only need poll() for PSI.
> +        *
> +        * See #726 on GitHub.
> +        */
> +       if (!opts->psi_poll_on) {
> +               fuse_pollhandle_destroy(ph);
> +               return -ENOSYS;
> +       }
> +
>         type = file_info_type(fi);
>  
>         if (LXCFS_TYPE_PROC(type)) {
> ```

Thanks @mihalicyn ! This works well too, would you like me to update the PR or do you want to push your own commit to it?

Thanks

mihalicyn:
> > Hi Tom,
> > thank you for the patch. WDYT about the following:
> > ```
> > diff --git a/src/lxcfs.c b/src/lxcfs.c
> > index 9ecdffd..5fb3360 100644
> > --- a/src/lxcfs.c
> > +++ b/src/lxcfs.c
> > @@ -540,9 +540,24 @@ int lxcfs_write(const char *path, const char *buf, size_t size, off_t offset,
> >  int lxcfs_poll(const char *path, struct fuse_file_info *fi,
> >                struct fuse_pollhandle *ph, unsigned *reventsp)
> >  {
> > +       struct fuse_context *fc = fuse_get_context();
> > +       struct lxcfs_opts *opts = fc->private_data;
> >         int ret;
> >         enum lxcfs_virt_t type;
> >  
> > +       /*
> > +        * Note, that if fuse daemon returns ENOSYS once, then
> > +        * kernel caches this result in fm->fc->no_poll and never
> > +        * sends FUSE_POLL again. It is not a problem for us now,
> > +        * because we only need poll() for PSI.
> > +        *
> > +        * See #726 on GitHub.
> > +        */
> > +       if (!opts->psi_poll_on) {
> > +               fuse_pollhandle_destroy(ph);
> > +               return -ENOSYS;
> > +       }
> > +
> >         type = file_info_type(fi);
> >  
> >         if (LXCFS_TYPE_PROC(type)) {
> > ```
> 
> Thanks @mihalicyn ! This works well too, would you like me to update the PR or do you want to push your own commit to it?
> 
> Thanks

yes, please update the PR. You don't need to mention me in the commit anyhow, it's just my review suggestion ;-)

tomponline:
I realised my local branch was `master` rather than `main` so have opened a new PR over here https://github.com/lxc/lxcfs/pull/727

